PR #2, opened 2018-07-19
Add the Garage class

leouieda:
Adds the `Garage` class and tests and utility functions that accompany it.
The class handles fetching data files from the remote url and loading in a registry from a file.
The download is done in streaming mode so large files are supported.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,45 +37,47 @@ TL;DR
     # Get the version string from your project. You have one of these, right?
     from . import __version__
 
+
     # Create a new garage to manage your sample data storage
     GARAGE = garage.create(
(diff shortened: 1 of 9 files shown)


leouieda:
@jrleeman this is the implementation of the `Garage` class that handles the downloading, checking the hash, and updating the file. I'd love some feedback if you have some time.

It doesn't handle the path parsing for the local folder or the environment variable. That will be delegated to a `garage.create` function that does the magic and then instantiates the `Garage` (see the README in this branch).

jrleeman:
Hey @leouieda - sorry we were out on an install. I should be able to play with this some this week though!

leouieda:
@jrleeman no problem. I opened #8 to keep track of progress.

PR #9, opened 2018-07-23
Add a registry builder utility

jrleeman:
Certainly open to any improvements on this, but with MetPy we have a ton of test data files, so building a registry by hand is out of the question. This will do so for us! Will need tests of course as well.

### DIFF
--- a/pooch/utils.py
+++ b/pooch/utils.py
@@ -1,6 +1,7 @@
 """
 Misc utilities
 """
+import os
 from pathlib import Path
 import sys
 import hashlib
@@ -9,7 +10,7 @@ from packaging.version import Version
 
 
 def os_cache(project, platform=None):
-    r"""
+    """
     Default cache location based on the operating system.
 
     Will insert the project name in the proper location of the path.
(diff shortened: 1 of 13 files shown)


leouieda:
@jrleeman looks great! `make_file_registry` should be a part of API so you can import it in `pooch/__init__.py` and add it to `docs/api/index.rst`. 

For tests, you could add a folder with a few more data files to [`pooch/tests/data`](https://github.com/fatiando/pooch/tree/master/pooch/tests/data). It would be good to add a section to `docs/usage.rst` to explain how to do this.

pooch/utils.py
```diff
@@ -138,3 +139,35 @@ def check_version(version, fallback="master"):
...
+    if recursive:
+        files = [os.path.join(dp, f) for dp, dn, fn in os.walk(os.path.expanduser(directory)) for f in fn]
+    else:
+        files = [f for f in os.listdir(directory) if os.path.isfile(os.path.join(directory, f))]
+
+    # Get the hash of each of the files
+    hashes = [file_hash(f) for f in files]
+
+    # Write out the files and hashes to the desired file
+    outfile = open(output, 'w')
```
leouieda: For testing, it would help to check if `isinstance(output, str)` before opening so we can pass in a StringIO. You can still use `outfile` in a `with` block to get rid of the `.close()`:

```
with outfile:
    for f, h in zip(files, hashes):
        outfile.write('{} {}\n'.format(f, h))
```

leouieda:
@jrleeman you'll want to rebase (or merge master). #12 fixed some issues with Python 3.5.

leouieda:
Hi @jrleeman, I'm moving this along to get this feature out. No pressure. I found some corner cases that I hand't thought about with files in subdirectories.

jrleeman:
Sorry! Things have been on fire with 3 more trips on the books by the end of Sept.

leouieda:
@jrleeman I think this is done. I'll merge it now and make another alpha release we can start testing on metpy.

Issue #8, opened 2018-07-20
Try using Pooch in MetPy


leouieda:
Use Pooch to replace the [MetPy](https://github.com/Unidata/MetPy) data caching system. This will help test the usefulness of loading a registry file from a file since they have a ton of data.

leouieda:
@jrleeman @dopplershift I made an alpha release so we can test this out. Please take a look to see if there is anything that you'd need for MetPy that is missing.

PS: Someone was squatting the Garage name on PyPI so I had to change it.

jrleeman:
Sounds good - We'll kick the tires on it!

leouieda:
Just noticed that MetPy will support 2.7 until next year. Should we try to support 2.7 for Pooch right now or would you rather wait until MetPy drops 2.7?

dopplershift:
I think we'd really like to use Pooch as our caching solution, so if it's not horrible to make work on Python 2.7, it'd be preferable to do so.

leouieda:
It's probably not horrible. I'll give it a try to see what happens.

leouieda:
Scratch that. I just tried it and it broke on the setup.py :) (there is no recursive glob in 2.7)

leouieda:
Most of the issues regard pathlib. I was using `expanduser` and `resolve` which are only available on 3.5+. So some of that code will need to be rewritten with `os`. I'd still like to keep using `pathlib.Path` but maybe the expansion can be done before creating the `Path`.

leouieda:
OK, it's done in #17. That was actually not too bad. The main issues were all the nice new pathlib features I was using. Reverting a lot of them back to `os.path` works. I'll cut a new release and we can finally take it for a spin :rocket:

PR #17, opened 2018-08-10
Add Python 2.7 support

leouieda:
Trying it out to see what breaks. Currently, it's mostly the use `Path.expanduser` and `resolve`.

This is required by metpy (#8)


leouieda:
Python 2.7 sucks.

PR #23, opened 2018-08-29
Check if data dir is writable and warn the user

### LINKED ISSUE #22: pooch.create should test for write access to directories and have a failback for failures
See realworld implementation problem with Unidata/MetPy#933

`pooch.create` can fail to write to whatever cache directory it is presented with.  A failback mechanism should exist, like what does with `matplotlib` when it attempts to find writable locations.  There is also a [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for this. 

```python
POOCH = pooch.create(
    path=pooch.os_cache('metpy'),
    base_url='https://github.com/Unidata/MetPy/raw/{version}/staticdata/',
    version='v' + __version__,
    version_dev='master',
    env='TEST_DATA_DIR')
```

## Full error message

```
File "/opt/miniconda3/envs/prod/lib/python3.6/site-packages/pooch/core.py", line 143, in create
   os.makedirs(versioned_path)
 File "/opt/miniconda3/envs/prod/lib/python3.6/os.py", line 210, in makedirs
     makedirs(head, mode, exist_ok)
  File "/opt/miniconda3/envs/prod/lib/python3.6/os.py", line 210, in makedirs
    makedirs(head, mode, exist_ok)
  File "/opt/miniconda3/envs/prod/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/usr/share/httpd/.cache'
```

thank you

leouieda:
`pooch.create` now only checks for a PermissionError when trying to
create the data directory and warns the user to set the environment
variable if the error occurs. Also checks if we can create a tempfile in
the directory if it already exists and issues the same warning.
The exception is not passed on so that modules that call `create` won't
break at import time. Instead, the failure will happen when
`Pooch.fetch` tries to create the directory or write to a file if it
exists.

Fixes #22

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -2,14 +2,20 @@
 Functions to download, verify, and update a sample dataset.
 """
 import os
+import sys
 from pathlib import Path
 import shutil
-from tempfile import NamedTemporaryFile
+import tempfile
 from warnings import warn
 
 import requests
 
 from .utils import file_hash, check_version
+
+
+# PermissionError was introduced in Python 3.3. This can be deleted when dropping 2.7
+if sys.version_info[0] < 3:
+    PermissionError = OSError  # pylint: disable=redefined-builtin,invalid-name
 
 
 def create(path, base_url, version, version_dev, env=None, registry=None):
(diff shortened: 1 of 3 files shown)


leouieda:
Waiting on conda-forge/pytest-feedstock#55 to fix the pytest issues on Mac.

Issue #4, opened 2018-07-20
Data in more than one remote source


leouieda:
Right now, we assume that all the sample data is in the same remote source. This might not be the case for many projects if they are using publicly available data. It would be great to be able to specify the a remote URL for individual entries in the registry (which should probably not be versioned). 

One of doing this that might not require much refactoring is to include a new attribute in `Pooch.registry_urls` that is a dictionary mapping file names to remote URLs where they can be downloaded.  We should not append the file name to this URL. The `Pooch._download_file` method can then check `if fname in self.registry_urls` and use `self.registry_urls[fname]` as the download source instead.

remram44:
This sounds like a good idea, but you might also want to isolate the part that resolves the file name into URL+hash to a specific method, so that it might be overridden in subclasses. Remember that this tool targets programmers, exposing a programming interface is probably a plus :wink:

In particular my remote might be using `so/me_file.csv` format (like Git) to limit the number of files in a single directory, download the registry itself, or do any number of other things; having to copy and edit pooch for that is probably unnecessary.

leouieda:
@remram44 thanks for the feedback! I agree that `Pooch` could use a refactoring with subclassing in mind. I created #27 to keep track of this. Any comments there would be very welcome :+1: 

> In particular my remote might be using so/me_file.csv format (like Git) to limit the number of files in a single directory, download the registry itself, or do any number of other things;

Not sure I understand. Could you clarify?

remram44:
This is a simple example of a use-case for custom name-to-URL code. If you are already convinced, you can ignore it :wink:

I described the scheme that is used by some software to avoid having too many files in a single flat folder (filesystems are slow in that situation), for example Git does this (look in your `.git/objects` folder, you'll find a bunch of folders with the first two characters of hashes, then files named with the rest of the hash).

leouieda:
Oh, yeah I get it now. Files in subdirectories of the cache are already supported. MetPy is using this feature: https://github.com/Unidata/MetPy/blob/master/metpy/io/tests/test_nexrad.py#L78

PR #31, opened 2018-10-27
Factor _get_url() for easy subclassing

remram44:
Addresses https://github.com/fatiando/pooch/issues/4#issuecomment-420761034

Without this, trying to add functionality in subclasses means having to override the entire `_download_file` method.


leouieda:
Thanks! I really appreciate it 🎉

PR #32, opened 2018-10-27
Correctly delete temporary downloads on error

remram44:
Because you create the temporary file with `delete=False`, it needs to be deleted manually if returning early (for example raising).

*(note to self: push follow-up PR after this)*


leouieda:
@remram44 thanks!

PR #33, opened 2018-10-30
Avoid copies across filesystem

remram44:
If the temporary directory is a separate filesystem (often the case), the `shutil.move()` call will actually need to copy the file contents, rather than being an O(1) operation.

Creating the file in the destination folder avoids this issue.


remram44:
Good call! Applied your suggestion, and now tests pass.

leouieda:
Perfect! Also, please feel free to add yourself to the `AUTHORS.md` file (if you want, of course).

PR #44, opened 2019-02-20
Follow XDG specification for cache location

### LINKED ISSUE #43: Use the XDG_CACHE_HOME environment variable if it exists
Following the discussion in #26, we need to make some changes to the `os_cache` function in [pooch/utils.py](https://github.com/fatiando/pooch/blob/master/pooch/utils.py) function to be compliant with the FreeDesktop standards:

* Add an option `check_environment` to the function that defaults to `False`
* If `check_environment == True`, check if the `XDG_CACHE_HOME` environment variable is defined.
* If it is, use it to define the default cache location instead of `~/.cache` on Linux.

remram44:
Fixes #43

### DIFF
--- a/pooch/utils.py
+++ b/pooch/utils.py
@@ -1,6 +1,7 @@
 """
 Misc utilities
 """
+import os
 from pathlib import Path
 import sys
 import hashlib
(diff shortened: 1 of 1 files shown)


leouieda:
Closing this since it's outdated compared to #45

PR #45, opened 2019-02-20
Use appdirs to get the cache directory

### LINKED ISSUE #43: Use the XDG_CACHE_HOME environment variable if it exists

remram44:
Rely on this is tiny module (no dependencies, one file, 600 lines) to do the right thing instead of implementing it ourselves. It follows the XDG specs for *NIX, but also correctly reads the location for Windows from the registry.

Adds `appdirs` as a dependency. Breaks compatibility by removing the `platform` argument of `os_cache` since it's no longer used. This should not be an issue since it was only there for testing purposes.

Fixes #43


pooch/utils.py
```diff
@@ -40,15 +41,7 @@ def os_cache(project, platform=None):
...
-    if platform is None:
-        platform = sys.platform
-    if platform == "darwin":
-        cache_path = Path("~", "Library", "Caches", project)
-    elif platform == "win32":
-        cache_path = Path("~", "AppData", "Local", project, "cache")
-    else:  # *NIX
-        cache_path = Path("~", ".cache", project)
-    return cache_path
+    return appdirs.user_cache_dir(project)
```
remram44: Should this be `Path(appdirs.user_cache_dir(...))`? The doctest seems to indicate so

leouieda:
@remram44 I'll make some changes to fix the failing tests on this and merge it in. Thanks for all your input on this stuff!

remram44:
You're very welcome!

pooch/utils.py
leouieda: It should. I changed that before merging.

leouieda:
:+1: It's always good to get rid of code :)

PR #49, opened 2019-03-21
Create private method to check if file is in registry

santisoler:
The file-in-registry check could be needed on future methods.
Instead of copy and paste the `if ... then raise ...` lines, would be better to have a single
private method that perform this task.
Also, it reduces the number of tests functions: instead of testing file-in-registry for each
new method, we can do it only once.


pooch/core.py
```diff
@@ -248,8 +248,7 @@ def fetch(self, fname):
             storage.
 
         """
-        if fname not in self.registry:
-            raise ValueError("File '{}' is not in the registry.".format(fname))
+        self._is_file_in_registry(fname)
```
leouieda:
```suggestion
        self._assert_file_in_registry(fname)
```

pooch/core.py
```diff
@@ -270,6 +269,10 @@ def fetch(self, fname):
             self._download_file(fname)
         return str(full_path)
 
+    def _is_file_in_registry(self, fname):
```
leouieda:
```suggestion
    def _assert_file_in_registry(self, fname):
        """
        Check if a file is in the registry and raise :class:`ValueError` if it's not.
        """
```

santisoler:
That's a better name for the method!
Merge when ready, then I can update #50 to make use of the `_assert_file_in_registry` method.

PR #59, opened 2019-03-30
Add post-download hook to Pooch.fetch

### LINKED ISSUE #35: Add mechanism for post download hooks
## Feedback or description of feature requested

It would be great to provide a mechanism to specify functions that are run after a file is downloaded (further checks, unpacking into a cache folder, etc). This could be implemented as a callback function that receives the `Pooch` object (`self`) and the file name that was downloaded. It would be called from `Pooch._download_file` after the download is finished.

leouieda:
Add a new argument `hook` to `Pooch.fetch` to pass in a function that is
executed right before returning and can overwrite the file path that is
returned to the user. Example usage would be to unzip an archive only
when downloading the file and return the path to the unzipped file. Will
be very useful to perform unpacking operations on larger files that can
be time consuming and we only want to do once.

Fixes #35

**Reminders**


doc/usage.rst
```diff
@@ -187,6 +187,76 @@ names of these files in the registry **must use Unix-style separators** (``'/'``
...
+           The full path to the unzipped file.
+           (Return the same fname is your hook doesn't modify the file).
+
+        """
+        # Create a new name for the unzipped file
+        unzipped = fname + ".unzipped"
+        # Don't unzip if file already exists and is not being downloaded
+        if action in ("update", "download") or not os.path.exists(unzipped):
+            with ZipFile(fname, "r") as zip_file:
+                # Extract the data file from within the archive
```
santisoler: Isn't better to use `ZipFile.extract` rather than `ZipFile.open`? For example, if the desired zipped files is inside a directory (within the zip file), I think the `open` method would raise an error due to missing unzipped directory, while the `extract` would create it.

leouieda: I was using it at first but the problem is that you can't set an output file name in `extract`, just the directory. So for this example case, this is the best approach.

> if the desired zipped files is inside a directory (within the zip file)

It shouldn't be a problem because we're not saving with that file name, we're using `unzipped` instead.

An alternative way if you have multiple files would be to set `unzipped` as the output directory and then use `extract`. We shouldn't extract into the data directory because there might be file name conflicts that are out of our control.

leouieda:
Thanks for the review and ideas :+1: 

> One idea is leave the code as it is and make the hook function to return a list of all the unzipped files.

That is perfectly valid with the current implementation. When you pass a hook, it will take over the return value of `fetch` and can return whatever it wants. My original idea for this was to return the directory containing the files but this would be just as good (or better).

> Another approach would be to allow to pass arguments to the hook function:

I don't think we need this. Since the user creates the hook function, they can pass in values as globals, default arguments, or by creating [partials with `functools`](https://docs.python.org/3.7/library/functools.html#functools.partial). 

> would be nice to add a multiple zipped files example to the docs

Yep, agreed. I'll add an example and a test case for this. I'll go with returning a list of file names to make sure it really works.

doc/usage.rst
santisoler: :+1:

santisoler:
> That is perfectly valid with the current implementation. When you pass a hook, it will take over the return value of `fetch` and can return whatever it wants. My original idea for this was to return the directory containing the files but this would be just as good (or better).

Yes, I was thinking that a list of unzipped files would be better, so the user don't have to remember each file name.

> I don't think we need this. Since the user creates the hook function, they can pass in values as globals, default arguments, or by creating [partials with `functools`](https://docs.python.org/3.7/library/functools.html#functools.partial).

That's a good approach.

> Yep, agreed. I'll add an example and a test case for this. I'll go with returning a list of file names to make sure it really works.

I'll wait for that example! I'm sure it will give a nice starting point for any user that needs to make us of this feature.

leouieda:
@santisoler this is ready for another round when you have the time

leouieda:
Thanks @santisoler. We should test this on rockhound before making a new release to make sure things work as expected.

PR #77, opened 2019-05-15
Adding untar processor

### LINKED ISSUE #76: Add an Untar processor

The `pooch.Unzip` processor unpacks a zip archive to a folder in the local file system after it's downloaded. We need a `pooch.Untar` processor that does the same thing but for `.tar` archives. The [tarfile](https://docs.python.org/3/library/tarfile.html) module can be used with the bonus that it supports compressed archives out of the box.

MatthewTurk:
This is a preliminary attempt at adding an untar processor.  If this looks like a good direction (especially with naming, etc) I would be more than happy to update it with comments -- and if it's not the direction that ought to be taken, I'll also happily close it!

I wasn't able to get the tests to work without modification since `tiny-data.tar.gz` isn't in the repository, but the tests do run if I manually specify the directory.

Fixes #76

**Reminders**


leouieda:
@MatthewTurk thanks, this is better than what I had in mind actually :)

The test data files are a bit of a chicken-and-egg problem in these cases. I added it to the master branch and updated your branch. Things should work now.

I don't think we need an entire section on this in the tutorial so all you need now is an entry in `doc/api/index.rst`.

pooch/processors.py
```diff
@@ -50,36 +52,113 @@ def __call__(self, fname, action, pooch):
...
+                tar_file.extractall(path=extract_dir)
+            else:
+                for member in self.members:
+                    warn(
+                        "Extracting '{}' from '{}' to '{}'".format(
+                            member, fname, extract_dir
+                        )
+                    )
+                    # Extract the data file from within the archive
+                    with tar_file.getmember(member) as data_file:
```
leouieda:
```suggestion
                    with tar_file.extractfile(member) as data_file:
```

I think `getmember` only returns metadata, not the actual contents. `extractfile` returns the member as a file-like object that we can `.read()` later.

pooch/processors.py
```diff
@@ -50,36 +52,113 @@ def __call__(self, fname, action, pooch):
...
+                        # Save it to our desired file name
+                        with open(os.path.join(extract_dir, member), "wb") as output:
+                            output.write(data_file.tobuf())
```
leouieda:
```suggestion
                            output.write(data_file.read())
```

pooch/processors.py
MatthewTurk: Ah!  OK, thank you -- and this makes it easier to then call `read`, as suggested below.

pooch/processors.py
MatthewTurk: Excellent, thanks!

leouieda:
Ah the joys of Python 2.7... It seems like the `data_file` returned by `extractfile` isn't a context manager in 2.7. I'll patch this up if that's alright with you.

leouieda:
I separated the tests into their own file (they shouldn't have been in `core` from the start) and added a few more checks to get the coverage to 100%. Hopefully all of the CIs will be satisfied this time :) 

Thanks for implementing this, Matt! I really appreciate it :1st_place_medal:

leouieda:
:confetti_ball:  Thanks for the contribution, Matt :confetti_ball:

MatthewTurk:
Hooray!  Thank you for the fixups!

PR #78, opened 2019-05-20
Add a decompression processor

### LINKED ISSUE #73: Add decompression processors

Many times, downloaded files will be compressed to save space in storage and during download. But it's sometimes a hassle to decompress them every time before reading the file. In the case of netCDF grids, decompressing into a temporary file makes it impossible to have `xarray` lazily load the grid. 

#59 introduced processor hooks and #72 introduced the `Unzip` processor. Following this logic, it would be good to have a `Decompress` processor that decompresses the downloaded file. It can guess the compression method from the file name or be given it explicitly as an `__init__` argument. 

The class would look something like this: 

```python
class Decompress:
    def __init__(self, method="auto"):
        self.method = method
    def __call__(self, fname, action, pooch):
        ...
        return decompressed
```

leouieda:
The `Decompress` processor saves a decompressed version of the downloaded file.
The new file will have `.decomp` added to the file name.
The compression method can be specified through the `method` argument and defaults to
`auto` (guessed based on the file extension).
Supports gzip, lzma/xz, and bzip2 through the Python builtin libraries.

Under Python 2.7, lzma and bzip2 require the `backports.lzma` and `bz2file` packages 
as well. These are soft dependencies and not required to use pooch.

Fixes #73

**Reminders**


leouieda:
Still needs docs.

PR #97, opened 2019-08-06
Add optional progress bar using tqdm

### LINKED ISSUE #91: Add progress bar for downloads

Add a progress bar for downloads using, say, [tqdm](https://github.com/tqdm/tqdm) or [python-progressbar](https://github.com/WoLpH/python-progressbar).
Getting this level of feedback is nice especially if you're on a slow connection.

Currently pooch issues a warning when something is being downloaded.
This is useful feedback because the dataset will be only be downloaded the first time you fetch it.
If there's more than one way to provide feedback, some changes to the API might be necessary to specify which one.

danshapero:
This patch adds a progress bar for downloads if the user has [tqdm](https://github.com/tqdm/tqdm) installed and passes the argument `progressbar=True` to the HTTPDownloader. Fixes #91.

I'm not sure what the best way is to go about testing an optional feature like this. Also, since the progress bar already gives feedback, maybe the warnings on download or unzipping could be removed?


pooch/downloaders.py
```diff
@@ -75,6 +81,7 @@ class HTTPDownloader:  # pylint: disable=too-few-public-methods
 
     def __init__(self, **kwargs):
         self.kwargs = kwargs
+        self.progressbar = self.kwargs.pop("progressbar", False) and has_tqdm
```
leouieda: Since we say in the docstring that kwargs are passed to `requests.get`, it's best if `progressbar=False` is an explicit argument here. It's probably also best to just assign the value to the attribute here `self.progressbar = progressbar` and check if `has_tqdm` in the `__call__` method.

leouieda: Another thing to consider adding here is warning in case `progress==True and has_tqdm==False` so the user knows why their progressbar isn't showing up.

pooch/downloaders.py
```diff
@@ -100,9 +107,22 @@ def __call__(self, url, output_file, pooch):
         try:
             response = requests.get(url, **kwargs)
             response.raise_for_status()
-            for chunk in response.iter_content(chunk_size=1024):
-                if chunk:
-                    output_file.write(chunk)
+            if self.progressbar:
```
leouieda:
```suggestion
            if self.progressbar and has_tqdm:
```

pooch/downloaders.py
```diff
@@ -100,9 +107,22 @@ def __call__(self, url, output_file, pooch):
...
+                total_size = int(response.headers.get("content-length", 0))
+                with tqdm.tqdm(
+                    response.iter_content(1024),
+                    total=total_size,
+                    unit="B",
+                    unit_scale=True,
+                ) as bar:
```
leouieda: This might be better broken in 2 lines:

```python
bar = tqdm.tqdm(...)
with bar:
    ...
```

pooch/downloaders.py
```diff
@@ -100,9 +107,22 @@ def __call__(self, url, output_file, pooch):
         try:
             response = requests.get(url, **kwargs)
             response.raise_for_status()
-            for chunk in response.iter_content(chunk_size=1024):
-                if chunk:
-                    output_file.write(chunk)
+            if self.progressbar:
+                total_size = int(response.headers.get("content-length", 0))
+                with tqdm.tqdm(
+                    response.iter_content(1024),
```
leouieda: Is this line needed if updating the bar manually in line 120?

pooch/downloaders.py
```diff
@@ -100,9 +107,22 @@ def __call__(self, url, output_file, pooch):
...
+                    total=total_size,
+                    unit="B",
+                    unit_scale=True,
+                ) as bar:
+                    for chunk in response.iter_content(chunk_size=1024):
+                        if chunk:
+                            bar.update(len(chunk))
+                            output_file.write(chunk)
+            else:
+                for chunk in response.iter_content(chunk_size=1024):
```
leouieda: Now that I see it repeated here, we should probably have `chunck_size` as a variable to avoid repetition.

danshapero:
I've factored out some of the repeated code, changed the initializer for HTTPDownloader, and added a warning in case the user asks for a progress bar but doesn't have tqdm.

How should the testing for this feature work? Since it's optional, do you want to test both with and without tqdm? Or just add it as a requirement?

leouieda:
@danshapero sorry for letting this sit for so long. I've been thinking if it's worth the hassle of setting up testing with and without tqdm since it's such a light weight dependency. It's pure Python and doesn't have any dependencies itself. So I think we can just add it as a required dependency but leave verbosity off by default. We can discuss whether this should be on by default on Pooch 1.0 later on and give time for projects that use Pooch to adapt.

leouieda:
So to answer your question, we can test with tqdm only. And you could also remove the `try` checks for tqdm installed.

danshapero:
My turn to apologize for taking forever :) I've made tqdm a hard requirement in setup.py and requirements.txt, added the progressbar to one of the core tests, and fixed various linter/style issues.

leouieda:
Thanks, @danshapero! I'm having a look at this and trying to think of a way to test the output of the progress bar (to see if it really is being printed correctly). I'm having a bit of trouble because the size of `tiny-data.txt` reported by requests seems be wrong. Add a larger test data file in #102 to test if this is the case. Will update you on the progress.

danshapero:
Yes I noticed the same thing. I tried it myself on a larger file and it all looked fine, but it's also worth testing for real.

leouieda:
@danshapero I think I have this working now. The trouble was not with the file size, but with the way text files are sent by Github. Apparently, they come gzip compressed and requests decompresses automatically. So updating the bar with `len(chunk)` didn't work because the decompressed chunk was much larger than the file size reported by requests. So I'm using `chunk_size` to update and manually completing the progress bar in the end. This works fine for both text and binary files. 

I added a test that captures stderr and checks that the progress bar is printed correctly. I also made tqdm optional by adding some extra builds to the CIs. Now let's see if all tests pass.

leouieda:
The Travis failure is unrelated and I'm fixing it separately. Will merge this in as soon as I have a fix for that.

danshapero:
Ok, do you want me to rebase out my silly style and linter fixes?

leouieda:
@danshapero no need. We squash and merge so the pr will be condensed into 1 commit anyway.

leouieda:
:confetti_ball: thanks @danshapero for this contribution! Sorry for taking over in the end there. This turned out to be a lot trickier than expected. I wanted to get this done quickly so we can include it in our 0.6.0 release.

danshapero:
@leouieda No worries, thanks for the help!

PR #100, opened 2019-10-06
Drop support for Python 2.7

hugovk:
Python 2.7 reaches EOL on 2020-01-01 and many projects are dropping support for it.

Fixes #99.

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [n/a] Add tests for new features or tests that would have caught the bug that you're fixing.
- [n/a] Add new public functions/methods/classes to `doc/api/index.rst`.
- [n/a] Write detailed docstrings for all functions/methods.
- [n/a] If adding new functionality, add an example to the docstring, gallery, and/or tutorials.

TODO from #99:

* [ ] Edit the conda-forge recipe to build noarch since we no longer need different dependencies depending on Python version.

I'm not sure when to do this.

PR #111, opened 2019-10-22
Add Hugo van Kemenade (hugovk) to author list

hugovk:
For [PR 💯](https://github.com/fatiando/pooch/pull/100)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -5,7 +5,7 @@ This project was initially created by [Leonardo Uieda](http://www.leouieda.com/)
 The following people have made contributions to the project (in alphabetical order by
 last name) and are considered "The Pooch Developers":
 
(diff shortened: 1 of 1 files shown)


leouieda:
@hugovk we've been keeping the list in alphabetical order by last name, so you should probably be first :slightly_smiling_face:

hugovk:
It depends on which country's surname sort order you use! But this being Python, I guess we can follow Guido's lead :)

https://github.com/python/cpython/blob/dfe726b1ace03f206f45253b93ed7610473ae20f/Misc/ACKS#L1416

Updated!

leouieda:
@hugovk Guido knows best :) Thanks!

leouieda:
> Dutch sort

:rofl:

hugovk:
🇳🇱

PR #115, opened 2019-10-25
Using logging API instead of warnings

### LINKED ISSUE #114: Use logging instead of warnings for user feedback

Pooch currently uses warnings to give feedback to users when a file is being downloaded, an archive unpacked, etc. The logging module from the Python standard library is a better fit for this purpose.

* More easily adjust how important the feedback is -- downloading a file is information, not having permission to create a destination directory is a warning or error
* Users can choose to silence feedback from pooch without silencing warnings from other libraries

danshapero:
This patch replaces the use of warnings to give feedback to the user
with the logging functionality from the Python standard library. Using
the logging API means that messages can be logged with different
priorities and the user can select which priority they want to see.
Fixes #114 .

**Reminders**

### DIFF
--- a/pooch/__init__.py
+++ b/pooch/__init__.py
@@ -2,7 +2,7 @@
 # Import functions/classes to make the API
 from . import version
 from .core import Pooch, create
-from .utils import os_cache, file_hash, make_registry, check_version
+from .utils import os_cache, file_hash, make_registry, check_version, get_logger
 from .downloaders import HTTPDownloader, FTPDownloader
 from .processors import Unzip, Untar, Decompress
 
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -10,7 +10,7 @@ from warnings import warn
 import ftplib
 
 import requests
-from .utils import file_hash, check_version, parse_url
+from .utils import file_hash, check_version, parse_url, get_logger
 from .downloaders import HTTPDownloader, FTPDownloader
 
 KNOWN_DOWNLOADERS = {
@@ -374,10 +374,20 @@ class Pooch:
 
         if action in ("download", "update"):
             action_word = dict(download="Downloading", update="Updating")
+
+            # Remove this if we switch to logging instead of warnings.
             warn(
                 "{} data file '{}' from remote data store '{}' to '{}'.".format(
                     action_word[action], fname, self.get_url(fname), str(self.path)
                 )
+            )
+
+            get_logger().info(
+                "%s data file '%s' from remote data store '%s' to '%s'.",
+                action_word[action],
+                fname,
+                self.get_url(fname),
+                str(self.path),
             )
 
             parsed_url = parse_url(url)
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -6,7 +6,6 @@ import os
 from pathlib import Path
 import shutil
 import tempfile
-from warnings import warn
 import ftplib
 
 import requests
@@ -172,17 +171,15 @@ def create(
             tempfile.NamedTemporaryFile(dir=path)
     except PermissionError:
         message = (
-            "Cannot write to data cache '{}'. "
-            "Will not be able to download remote data files. ".format(path)
+            "Cannot write to data cache '%s'. "
+            "Will not be able to download remote data files. "
         )
+        args = [path]
         if env is not None:
-            message = (
-                message
-                + "Use environment variable '{}' to specify another directory.".format(
-                    env
-                )
-            )
-        warn(message)
+            message += "Use environment variable '%s' to specify another directory."
+            args += [env]
+
+        get_logger().warning(message, *args)
     pup = Pooch(path=Path(path), base_url=base_url, registry=registry, urls=urls)
     return pup
 
@@ -374,14 +371,6 @@ class Pooch:
 
         if action in ("download", "update"):
             action_word = dict(download="Downloading", update="Updating")
-
-            # Remove this if we switch to logging instead of warnings.
-            warn(
-                "{} data file '{}' from remote data store '{}' to '{}'.".format(
-                    action_word[action], fname, self.get_url(fname), str(self.path)
-                )
-            )
-
             get_logger().info(
                 "%s data file '%s' from remote data store '%s' to '%s'.",
                 action_word[action],
--- a/pooch/tests/test_core.py
+++ b/pooch/tests/test_core.py
@@ -6,6 +6,8 @@ import sys
 from pathlib import Path
 from tempfile import TemporaryDirectory
 import tempfile
+from io import StringIO
+from logging import StreamHandler
 
 import pytest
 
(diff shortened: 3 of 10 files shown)


danshapero:
It looks like the tests are failing on Windows but not on Mac or Linux; I don't have a Windows machine to test on. I think there might be some difference in how temporary files work. Any suggestions?

leouieda:
@danshapero yep, we had a lot of failures before on Windows because of temporary file management. Be sure to use `with` blocks for your temp files so they are always closed and cleaned up. That usually helps.

I also see that there are a lot of changes in this PR. It might be best to focus on replacing warnings with logs in a single place so we can decide on an API. After that, you can propagate the changes to other parts of the code. That will make this a bit easier to review. Sorry for not specifying this earlier.

danshapero:
You're right, this was a huge change, and what with other recent improvements that you made it was difficult for me to rebase from master. I've reduced it down to a smaller change so it'll be easier to review and iterate on. I've added a logging event to the function `pooch.core.Pooch.fetch` and a test that the log is written properly to the `test_pooch_custom_url`.

danshapero:
I managed to fix the permission problem on Windows by logging to memory using [StringIO](https://docs.python.org/3/library/io.html#io.StringIO) rather than to a file on disk.

pooch/tests/test_core.py
```diff
@@ -53,18 +55,22 @@ def test_pooch_custom_url():
...
-            assert str(warn[-1].message).split()[-1] == "'{}'.".format(path)
+        # Create an in-memory file and a logging handler that writes events to
+        # this file, and check that downloading data gets logged
+        log_file = io.StringIO()
+        handler = logging.StreamHandler(log_file)
+        get_logger().addHandler(handler)
+        fname = pup.fetch("tiny-data.txt")
+        logs = log_file.getvalue()
+        assert logs.split()[0] == "Downloading"
+        assert logs.split()[-1] == "'{}'.".format(path)
```
leouieda: This would be great as a context manager:

```python
with capture_log() as log:
    fname = pup.fetch("tiny-data.txt")
    logs = log.getvalue()
    assert logs.split()[0] == "Downloading"
    assert logs.split()[-1] == "'{}'.".format(path)
```

The context would setup the handler and return the `StringIO`. When exiting it could do any clean up that it needs (remove the handler so it won't affect other tests down the line?). 

An easy way make a context manager is with [`contextlib`](https://docs.python.org/3.8/library/contextlib.html):

```python
from contextlib import contextmanager

@contextmanager
def capture_log():        
    log_file = io.StringIO()
    handler = logging.StreamHandler(log_file)
    get_logger().addHandler(handler)
    # return this to the context and halt until exit
    yield log_file
    # executed on exit
    get_logger().removeHandler(handler)
```

which could be added to `pooch/tests/utils.py`.

pooch/tests/test_core.py
```diff
@@ -53,18 +55,22 @@ def test_pooch_custom_url():
...
         check_tiny_data(fname)
-        # Check that no warnings happen when not downloading
-        with warnings.catch_warnings(record=True) as warn:
-            fname = pup.fetch("tiny-data.txt")
-            assert not warn
+        # Check that no logging happens when there are no events
+        fname = pup.fetch("tiny-data.txt")
+        new_logs = log_file.getvalue()
+        assert new_logs == logs
```
leouieda: With the context manager above, this part could become:

```python
with capture_log() as log:
    fname = pup.fetch("tiny-data.txt")
    assert not log.getvalue()
```

pooch/core.py
```diff
@@ -374,12 +374,22 @@ def mydownloader(url, output_file, pooch):
...
                     action_word[action], fname, self.get_url(fname), str(self.path)
                 )
             )
 
+            get_logger().info(
+                "%s data file '%s' from remote data store '%s' to '%s'.",
+                action_word[action],
+                fname,
+                self.get_url(fname),
+                str(self.path),
```
leouieda: To make life easier, you can probably just keep the string formatting from the original warnings. I don't think `info` requires the arguments and would be just as happy to get a pure string as input. Then it's only a matter of replacing `warn` with `get_logger().info`.

danshapero: PyLint gives a warning if you invoke `format` yourself on the log message (see [this SO answer](https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation)), which is why I've written it the way I have. The supposed advantage of the way I've written it, and the reason that PyLint gives a warning if you don't, is that the string interpolation is lazy. If the user only wants to log critical errors, then you won't waste time formatting the strings at, say, the "info" level.

Personally, I don't like this argument -- why do string formatting differently in a few places for a likely negligible performance gain? I'd rather write it the way you suggest, but then we'd have to disable `logging-format-interpolation` in the project's PyLint configuration.

leouieda: Right, that makes sense if the app is logging quite a bit of information. Pooch doesn't log much but maybe it could be an issue in people want to download vast quantities of small files (probably not, but who knows). Might be best to just stick with what pylint suggests.

pooch/tests/test_core.py
danshapero: Ooh this is a big improvement!

leouieda: :+1: Looks great!

leouieda:
@danshapero this looks great to me! I think it's safe to go ahead and make the full transition from warnings to logs.

danshapero:
@leouieda This is ready for another look. The test that nothing gets logged when you set the level to critical is a little weird -- pytest also captures logging events, so setting the log level within a test doesn't do what it normally does.

pooch/__init__.py
```diff
@@ -2,7 +2,7 @@
 # Import functions/classes to make the API
 from . import version
 from .core import Pooch, create
-from .utils import os_cache, file_hash, make_registry, check_version
+from .utils import os_cache, file_hash, make_registry, check_version, get_logger
```
leouieda: I don't think we need to expose `get_logger` to users since it's only used internally. In that case, it doesn't need to be here.

pooch/tests/test_core.py
```diff
@@ -74,21 +75,45 @@ def test_pooch_download():
...
+        critical_log_file = StringIO()
+        critical_handler = StreamHandler(critical_log_file)
+        critical_handler.setLevel("CRITICAL")
+        get_logger().addHandler(critical_handler)
+        path = Path(local_store)
+        urls = {"tiny-data.txt": BASEURL + "tiny-data.txt"}
+        # Setup a pooch in a temp dir
+        pup = Pooch(path=path, base_url="", registry=REGISTRY, urls=urls)
+        # Check that the logs say that the file is being downloaded
+        with capture_log() as log_file:
```
leouieda: Would it possible to pass in a `level` argument to `capture_log` and have it set the level on the handler?

pooch/__init__.py
danshapero: Ok, I added this for users who want to change the logging level -- something like `pooch.get_logger().setLevel("CRITICAL")`, since you can just pass a string for the level. I can remove it from the public API but I think it might be good to have another way to set the logging level then.

leouieda: Ah right, that makes perfect sense. We can leave it then.

pooch/tests/test_core.py
danshapero: Very possible and much better, thanks!

leouieda:
@danshapero this looks great! Could you add a little section in `doc/usage.rst` with an example of silencing Pooch in case someone doesn't want the verbosity? After that, we should be good to go.

leouieda:
Might be worth merging in `master` before continuing.

danshapero:
I wrote enough documentation to explain how to silence the logging events, but I can change it as you see fit.

leouieda:
Awesome @danshapero! Looks great to me. Merging!

PR #117, opened 2019-11-04
Add support for file objects to load_registry()

remram44:
This makes it possible to pass a file object to `load_registry()`, opened either in binary or text mode.

My main reason for this is to support passing `pkg_resources.resource_stream()` to it, which is the preferred way of using data files from the package (it will allow your package to be "zip-safe" though that doesn't matter as much those days).

This could be updated:
https://github.com/fatiando/pooch/blob/9a1382d68df2345f6120fd36456537a8a495e1db/doc/usage.rst#L461
to this:
```python
GOODBOY.load_registry(pkg_resources.resource_stream("plumbus", "registry.txt"))
```

**Reminders**

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -1,6 +1,7 @@
 """
 The main Pooch class and a factory function for it.
 """
+import contextlib
 import os
 from pathlib import Path
 import shutil
@@ -440,12 +441,22 @@ class Pooch:
 
         Parameters
         ----------
-        fname : str
-            File name and path to the registry file.
+        fname : str | fileobj
+            File path to the registry file, of file object.
 
         """
-        with open(fname) as fin:
+        with contextlib.ExitStack() as stack:
+            if hasattr(fname, "read"):
+                # It's a file object
+                fin = fname
+            else:
+                # It's a file path
+                fin = stack.enter_context(open(fname))
+
             for linenum, line in enumerate(fin):
+                if isinstance(line, bytes):
+                    line = line.decode("utf-8")
+
                 elements = line.strip().split()
                 if len(elements) > 3 or len(elements) < 2:
                     raise OSError(
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -442,7 +442,7 @@ class Pooch:
         Parameters
         ----------
         fname : str | fileobj
-            File path to the registry file, of file object.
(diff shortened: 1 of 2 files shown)


remram44:
Alternatively there could be a separate method that does the `load_registry(resource_stream(...))` in one go.

remram44:
(fixed lint issues: `'` -> `"`, `fp` -> `fin`)

pooch/core.py
```diff
@@ -440,12 +441,22 @@ def load_registry(self, fname):
 
         Parameters
         ----------
-        fname : str
-            File name and path to the registry file.
+        fname : str | fileobj
+            File path to the registry file, of file object.
```
leouieda:
```suggestion
            Path (or open file object) to the registry file.
```

leouieda: This is a tricky phrasing. Does this make it less ambiguous?

pooch/core.py
```diff
@@ -440,12 +441,22 @@ def load_registry(self, fname):
...
 
         """
-        with open(fname) as fin:
+        with contextlib.ExitStack() as stack:
```
leouieda: Ooh, this is new to me. Bookmarking for later use :slightly_smiling_face:

pooch/core.py
remram44: Ah yes, thanks

remram44:
> We can add the `pkg_resources` line to the docs but it might be better to include it as an alternative or recommended way.

Are you asking for it to be alternative or recommended?

To be honest it is pretty weird to *not* use `pkg_resources` in this context, since the example shows creating a complete package, with a setup.py and `package_data`. There are also multiple ways in which the `os.path` code can fail, for example if you move it to a sub-package, it will break unless you update it with an extra `dirname()` (or if running from a ZIP, os.path will construct paths inside the ZIP).

leouieda:
> Are you asking for it to be alternative or recommended?

Sorry, my comment wasn't clear. We should definitely recommend this as the preferred approach since it's the correct thing to do (we should probably do it internally as well :). But we should probably acknowledge that the other approach exists and why it's best to do it this way. This can be done in another PR as it's not directly related to this.

Merging this one in since it seems to be complete and all tests pass. Thanks! I always learn some new things from your PRs :slightly_smiling_face:

PR #120, opened 2019-11-07
Update doc to use pkg_resources

remram44:
Follow up on #117

Using pkg_resources instead of `os.path` calls is less error-prone and allows one's package to work from ZIPs (zip-safe).

**Reminders**

- Add tests for new features or tests that would have caught the bug that you're fixing.
- Add new public functions/methods/classes to `doc/api/index.rst`.
- Write detailed docstrings for all functions/methods.
- If adding new functionality, add an example to the docstring, gallery, and/or tutorials.


remram44:
I expect some bike-shedding to happen here, I'm really not sure what's the best way to present this.

doc/usage.rst
```diff
@@ -458,10 +459,15 @@ use :meth:`pooch.Pooch.load_registry` to read them:
...
-module and should be shipped with the package. It's contents are:
+    # Get registry file from package_data
+    registry_file = pkg_resources.resource_stream("plumbus", "registry.txt")
+    # Load this registry file
+    GOODBOY.load_registry(registry_file)
+
+The ``registry.txt`` file in this case is in the ``plumbus/`` directory module and should be
+shipped with the package. We use `pkg_resources <https://setuptools.readthedocs.io/en/latest/pkg_resources.html#basic-resource-access)`__
+to access it, giving it the name of our Python package (the one we'll put in ``setup.py``).
+Its contents are:
```
remram44: I wasn't sure what width is used for word-wrapping rst files. Also this link is too long :sweat_smile:

leouieda: :smile: for rst files it doesn't really matter too much. I usually do 79 or 88 (whichever you prefer). Not much we can do about the link, though.

doc/usage.rst
```diff
@@ -458,10 +459,15 @@ use :meth:`pooch.Pooch.load_registry` to read them:
...
-
-The ``registry.txt`` file in this case is in the same directory as the ``datasets.py``
-module and should be shipped with the package. It's contents are:
+    # Get registry file from package_data
+    registry_file = pkg_resources.resource_stream("plumbus", "registry.txt")
+    # Load this registry file
+    GOODBOY.load_registry(registry_file)
+
+The ``registry.txt`` file in this case is in the ``plumbus/`` directory module and should be
+shipped with the package. We use `pkg_resources <https://setuptools.readthedocs.io/en/latest/pkg_resources.html#basic-resource-access)`__
```
leouieda: Maybe instead of "We use package_resources" you could say: 

> We access the file contents using pkg_resources (instead of `os.path.dirname(__file__)`) because it's safer alternative. For example, it makes sure our code can work out of a zip archive, which is necessary depending on how the user installs the package.

doc/usage.rst
```diff
@@ -458,10 +459,15 @@ use :meth:`pooch.Pooch.load_registry` to read them:
...
-    GOODBOY.load_registry(os.path.join(os.path.dirname(__file__), "registry.txt"))
-
-The ``registry.txt`` file in this case is in the same directory as the ``datasets.py``
-module and should be shipped with the package. It's contents are:
+    # Get registry file from package_data
+    registry_file = pkg_resources.resource_stream("plumbus", "registry.txt")
+    # Load this registry file
+    GOODBOY.load_registry(registry_file)
+
+The ``registry.txt`` file in this case is in the ``plumbus/`` directory module and should be
```
leouieda:
```suggestion
The ``registry.txt`` file is in the ``plumbus/`` directory (the package) and should be
```

remram44: this repeats "the package"

leouieda: Right, didn't see that one in the next line. Just feel like "directory module" sounds strange.

leouieda:
@remram44 I pushed a fix for the broken hyperlink and removed the repeated "package". Looks good to me now and will merge as soon as CIs are happy. Thank you for this! :+1:

PR #113, opened 2019-10-24
Add support and CI testing for Python 3.8

hugovk:
Released on 2019-10-14 🎉 

* https://discuss.python.org/t/python-3-8-0-is-now-available/2478

Not yet available on Azure Pipelines:

* https://github.com/microsoft/azure-pipelines-image-generation/issues/1317

Nor Conda, so "Linux - Python 3.8" is left out from Travis CI for now:

* https://github.com/conda-forge/python-feedstock/pull/274

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [n/a] Add tests for new features or tests that would have caught the bug that you're fixing.
- [n/a] Add new public functions/methods/classes to `doc/api/index.rst`.
- [n/a] Write detailed docstrings for all functions/methods.
- [n/a] If adding new functionality, add an example to the docstring, gallery, and/or tutorials.

PR #122, opened 2019-11-13
Use napoleon instead of numpydoc to unpin sphinx

leouieda:
The combination of latest sphinx (2.2.0), numpydoc, and sphinx_rtd_theme
causes a mangled rendering of the parameter list. We were pinning sphinx
to 1.8.5, which is getting old and causing dependency problems.
Replacing numpydoc with sphinx.ext.napoleon fixes the problem (though
generates slightly different output). Needed to update the class
autosummary template to account for missing methods summary block.

See discussion in #113 

**Reminders**


leouieda:
This is what the update looks like:

## New class docs

class-new

## Current class docs

class-current

## New function docs

func-new

## Current function docs

func-current

leouieda:
:+1: we should apply this to the other Fatiando projects as well

PR #123, opened 2019-11-14
Make docstrings 79 characters (check with flake8)

### LINKED ISSUE #121: Wrap all docstrings to 79 characters

After fatiando/verde#177 raised by @prisae, fatiando/contributing#9 and fatiando/contributing#10 it has been decided that all docstrings must be wrapped to 79 characters per line.
We should warp all the existing docstrings in Pooch to 79 characters and also configure flake8 to check for this when running the style checks.

Sadly, there's no way (yet) to automatically change all docstrings, but at least we can use flake8 to raise lines that fail. This can be done by setting `max-doc-length` to 79 characters on the flake8 configuration under `setup.cfg`.

leouieda:
Docstrings with 88 characters (the Black line length) don't show up
nicely in Jupyter and IPython (since the display is 80 chars). This
means that our docstrings were unreadable.
See fatiando/verde#177 fatiando/contributing#9 fatiando/contributing#10
for context.
Make flake8 check that docstrings (and comments, sadly) are 79
characters so we can make sure we don't do it again accidentally.
Have to check comments as well because the flake8 flag doesn't
discriminate between them.

Fixes #121 

**Reminders**

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -24,52 +24,56 @@ def create(
     urls=None,
 ):
     """
-    Create a new :class:`~pooch.Pooch` with sensible defaults to fetch data files.
+    Create a :class:`~pooch.Pooch` with sensible defaults to fetch data files.
 
-    If a version string is given, the Pooch will be versioned, meaning that the local
-    storage folder and the base URL depend on the project version. This is necessary
-    if your users have multiple versions of your library installed (using virtual
-    environments) and you updated the data files between versions. Otherwise, every time
-    a user switches environments would trigger a re-download of the data. The version
-    string will be appended to the local storage path (for example,
-    ``~/.mypooch/cache/v0.1``) and inserted into the base URL (for example,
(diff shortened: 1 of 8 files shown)


leouieda:
Sorry @danshapero, I pinged you here by mistake. But please feel free to have a look as well :)

danshapero:
I'm a luddite and still use the terminal, so I think this is a good idea :)

leouieda:
> What do you think about this? Should we keep a max-doc-length to 79 on every file or just on the API?

I just find it simpler to have exceptions to rules :) It doesn't matter much since the tests aren't part of the API and docstrings are read in the code. In Pooch, it was relatively easy to change the test docstrings so I didn't make them an exception.

As for the examples, it also doesn't matter much since they are rendered by sphinx-gallery. So I'm fine with them being 88 instead of 79. That said, I'll probably write them as 79 simply because it's easier to not configure exceptions in my editor :)

PR #133, opened 2020-01-09
Allow to use different hashing algorithms

Xarthisius:
### Rationale
I'm trying to use pooch to manage/fetch data stored in external data repositories. All of them provide some sort of checksum, unfortunately it's not always SHA256. Ideally I'd like to generate pooch's registry file using hashes obtained directly from a given data repository. However, I need to be able to use proper hashing algorithm for data verification.

### Implementation
Extend registry format from:

```
name hash url
```

to 

```
name alg:hash url
```
Whenever old format registry is detected (i.e. `:` is not present in *hash*) `sha256:` is prepended to the hash automatically, which ensures backward compatibility.

### TODO

- [x] Update docs

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -377,8 +377,7 @@ class Pooch:
         self._assert_file_in_registry(fname)
 
         # Create the local data directory if it doesn't already exist
-        if not self.abspath.exists():
-            os.makedirs(str(self.abspath))
+        os.makedirs(str(self.abspath), exist_ok=True)
 
         full_path = self.abspath / fname
         url = self.get_url(fname)
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -365,16 +365,18 @@ class Pooch:
         full_path = self.abspath / fname
         url = self.get_url(fname)
         in_storage = full_path.exists()
-        hash_alg = self.registry[fname].split(":")[0]
+
         if not in_storage:
             action = "download"
-        elif (
-            in_storage
-            and file_hash(str(full_path), alg=hash_alg) != self.registry[fname]
-        ):
-            action = "update"
         else:
-            action = "fetch"
+            hash_alg = self.hash_algorithm(fname)
+            current_hash = "{}:{}".format(
+                hash_alg, file_hash(str(full_path), alg=hash_alg)
+            )
+            if current_hash != self.registry[fname]:
+                action = "update"
+            else:
+                action = "fetch"
 
         if action in ("download", "update"):
             action_word = dict(download="Downloading", update="Updating")
@@ -404,7 +406,7 @@ class Pooch:
             tmp.close()
             try:
                 downloader(url, tmp.name, self)
-                self._check_download_hash(fname, tmp.name, alg=hash_alg)
+                self._check_download_hash(fname, tmp.name)
                 # Ensure the parent directory exists in case the file is in a
                 # subdirectory. Otherwise, move will cause an error.
                 if not os.path.exists(str(full_path.parent)):
@@ -441,7 +443,7 @@ class Pooch:
         self._assert_file_in_registry(fname)
         return self.urls.get(fname, "".join([self.base_url, fname]))
 
-    def _check_download_hash(self, fname, downloaded, alg="sha256"):
+    def _check_download_hash(self, fname, downloaded):
         """
         Check the hash of the downloaded file against the one in the registry.
 
@@ -458,7 +460,10 @@ class Pooch:
             If the hashes don't match.
 
         """
-        tmphash = file_hash(downloaded, alg=alg)
+        registry_hash_alg = self.hash_algorithm(fname)
+        tmphash = "{}:{}".format(
+            registry_hash_alg, file_hash(downloaded, alg=registry_hash_alg)
+        )
         if tmphash != self.registry[fname]:
             raise ValueError(
                 "Hash of downloaded file '{}' doesn't match the entry in the registry."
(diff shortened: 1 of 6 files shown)


pooch/core.py
```diff
@@ -435,7 +442,7 @@ def get_url(self, fname):
         self._assert_file_in_registry(fname)
         return self.urls.get(fname, "".join([self.base_url, fname]))
 
-    def _check_download_hash(self, fname, downloaded):
+    def _check_download_hash(self, fname, downloaded, alg="sha256"):
```
leouieda: Is the default here needed since `hash_alg` should be set already if the code in `__init__` is correct?

pooch/core.py
```diff
@@ -467,7 +474,7 @@ def load_registry(self, fname):
 
         Use this if you are managing many files.
 
-        Each line of the file should have file name and its SHA256 hash
+        Each line of the file should have file name and its hash ("alg:checksum")
         separate by a space. Only one file per line is allowed. Custom download
         URLs for individual files can be specified as a third element on the
         line.
```
leouieda: This would also be a good place to say that the `alg:checkum` syntax is optional and sha256 will be used by default.

pooch/tests/utils.py
```diff
@@ -63,15 +63,18 @@ def pooch_test_registry():
...
+        "tiny-data.zip": "sha256:0d49e94f07bc1866ec57e7fd1b93a351fba36842ec9b13dd50bf94e8dfa35cbb",
+        "store.zip": "sha256:0498d2a001e71051bbd2acd2346f38da7cbd345a633cb7bf0f8a20938714b51a",
+        "tiny-data.tar.gz": (
+            "sha512:5f495fb71ba1844a1c2d6bd237fd8c9078536081a263059003f97d89714e"
+            "7bd1f376d17ecdaaa275dd887d53c27903e83493b2b1fb47350d00937ae0e96dfceb"
+        ),
+        "store.tar.gz": "sha256:088c7f4e0f1859b1c769bb6065de24376f366374817ede8691a6ac2e49f29511",
+        "tiny-data.txt.bz2": "sha256:753663687a4040c90c8578061867d1df623e6aa8011c870a5dbd88ee3c82e306",
+        "tiny-data.txt.gz": "sha256:2e2da6161291657617c32192dba95635706af80c6e7335750812907b58fd4b52",
+        "tiny-data.txt.xz": "sha256:99dcb5c32a6e916344bacb4badcbc2f2b6ee196977d1d8187610c21e7e607765",
```
leouieda: In the tests, it might best to keep the old syntax so we are certain that it's backwards compatible. We can then add a separate test for the new syntax. It's fine to use only `alg:checksum` in the documentation since we want people to use it in the future.

pooch/utils.py
```diff
@@ -59,20 +61,22 @@ def file_hash(fname):
...
+    if alg not in hashlib.algorithms_available:
+        raise ValueError("Algorithm '{}' not available in hashlib".format(alg))
+    hasher = hashlib.new(alg)
     with open(fname, "rb") as fin:
         buff = fin.read(chunksize)
         while buff:
             hasher.update(buff)
             buff = fin.read(chunksize)
-    return hasher.hexdigest()
+    return "{}:{}".format(alg, hasher.hexdigest())
```
leouieda: This is the only part that I'm not very sure what to do. This technically breaks compatibility since `file_hash` is in the public API. If possible, I'd prefer to keep this function returning only the hash to be on the safe side.

pooch/core.py
```diff
@@ -365,9 +368,13 @@ def mydownloader(url, output_file, pooch):
         full_path = self.abspath / fname
         url = self.get_url(fname)
         in_storage = full_path.exists()
+        hash_alg, _ = self.registry[fname].split(":")
         if not in_storage:
             action = "download"
-        elif in_storage and file_hash(str(full_path)) != self.registry[fname]:
+        elif (
+            in_storage
+            and file_hash(str(full_path), alg=hash_alg) != self.registry[fname]
```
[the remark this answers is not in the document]
Xarthisius: This one is tricky to do, cause that `file_hash` makes sense only if `in_storage` is True. I'd need to check that prior to defining `hash_changed` or make a nested if.

pooch/tests/utils.py
Xarthisius: Fixed in 0a5e712

pooch/utils.py
Xarthisius: Fixed in 0a5e712

pooch/core.py
Xarthisius: Fixed in 0a5e712

pooch/core.py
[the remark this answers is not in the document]
leouieda: Ah right, I forgot about that. It's fine the way it is then.

pooch/tests/utils.py
```diff
@@ -63,18 +63,22 @@ def pooch_test_registry():
...
+        "tiny-data.tar.gz": "41503f083814f43a01a8e9a30c28d7a9fe96839a99727a7fdd0acf7cd5bab63b",
+        "store.tar.gz": "088c7f4e0f1859b1c769bb6065de24376f366374817ede8691a6ac2e49f29511",
+        "tiny-data.txt.bz2": "753663687a4040c90c8578061867d1df623e6aa8011c870a5dbd88ee3c82e306",
+        "tiny-data.txt.gz": "2e2da6161291657617c32192dba95635706af80c6e7335750812907b58fd4b52",
+        "tiny-data.txt.xz": "99dcb5c32a6e916344bacb4badcbc2f2b6ee196977d1d8187610c21e7e607765",
     }
     return registry
+
+
+def add_hash_algs(registry):
```
leouieda: Would this be a useful function to have in `pooch/core.py`? The same code is used in `Pooch.__init__`.

leouieda:
@Xarthisius thanks for making those changes. Looks like this is almost good to go! Only need some text in the docstrings about specifying the hash algorithms and that one comment about the `add_hash_algs` function. 

If you have the time, it would also be nice to have a short section in `doc/usage.rst` page about specifying different hashes.

leouieda:
It would also be nice to have at least 1 test using a different hash algorithm.

Xarthisius:
@leouieda I believe I addressed all the comments. I'd appreciate a 2nd review.

leouieda:
@Xarthisius looks good to me :+1: Merging this in!

Thank you for the contribution to Pooch! It's much appreciated. Please feel free to open another PR adding yourself to the AUTHORS.md file (if you want, of course).

PR #134, opened 2020-01-13
Expand the section on Intake and include fsspec

leouieda:
As pointed out by reviewer @martindurant, we failed to mention fsspec in the "state of the field" section and could have done a better job at presenting Intake. Include a couple of sentences more on Intake and fsspec, how fsspec could be used in Pooch in the future, and what are the advantages and disadvantages of Pooch with respect to these tools.

openjournals/joss-reviews#1943

### DIFF
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -137,7 +137,7 @@ tar) and decompressing files (gzip, lzma, and bzip2).
 
 To the best of the authors' awareness, the only other Python software with some
 overlapping functionality are [Intake](https://github.com/intake/intake) and
-[fsspec](https://github.com/intake/filesystem_spec) (which is the backend for Intake).
+[fsspec](https://github.com/intake/filesystem_spec) (which is used by Intake).
 The fsspec library provides a unified interface for defining file systems and
 opening files, regardless of where the files are located (local system,
 HTTPS/FTP servers, Amazon S3, Google Cloud Storage, etc).
@@ -146,15 +146,13 @@ the one in Pooch, but has a wider range of download methods available.
 In the future, fsspec could be used as a backend to expand Pooch's download
 capabilities beyond HTTPS and FTP.
 Intake manages data catalogues (with download locations and extensive
-metadata), data download and caching (using fsspec), data loading,
-visualization, and browsing.
+metadata), data download and caching, data loading, visualization, and
+browsing.
 It has built-in capabilities for loading data into standard containers,
 including NumPy, pandas, and xarray.
-While Intake and fsspec are powerful and highly configurable,
-we argue that Pooch's advantage is in its simplicity.
-Pooch has a simpler setup and is more accessible to developers without advanced
-programming knowledge, meeting the immediate needs of busy scientific software
-authors and individual scientists.
+While Intake and fsspec are powerful and highly configurable tools,
+we argue that Pooch's strong points are its simplicity, straight-forward
+documentation, and focus on solving a single problem.
 
 The Pooch API is stable and has been field-tested by other projects:
 MetPy [@metpy], Verde [@verde], RockHound [@rockhound], predictatops
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -149,7 +149,7 @@ Intake manages data catalogues (with download locations and extensive
 metadata), data download and caching (using fsspec), data loading,
 visualization, and browsing.
 It has built-in capabilities for loading data into standard containers,
-including numpy, pandas, and xarray.
+including NumPy, pandas, and xarray.
 While Intake and fsspec are powerful and highly configurable,
 we argue that Pooch's advantage is in its simplicity.
 Pooch has a simpler setup and is more accessible to developers without advanced
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -136,13 +136,25 @@ authentication) as well as processing functions for unpacking archives (zip or
 tar) and decompressing files (gzip, lzma, and bzip2).
 
 To the best of the authors' awareness, the only other Python software with some
-overlapping functionality is [Intake](https://github.com/intake/intake).
-While Intake is powerful and can be used to manage large data archives,
-we argue that Pooch has a simpler setup and meets the
-specific needs of scientific software authors and individual scientists.
-For example, Pooch does not require users to change their data loading code to
-fit into a plug-in structure, instead only providing the file path for the
-user.
+overlapping functionality are [Intake](https://github.com/intake/intake) and
+[fsspec](https://github.com/intake/filesystem_spec) (which is the backend for Intake).
+The fsspec library provides a unified interface for defining file systems and
+opening files, regardless of where the files are located (local system,
+HTTPS/FTP servers, Amazon S3, Google Cloud Storage, etc).
+fsspec implements similar download and caching functionality to
+the one in Pooch, but has a wider range of download methods available.
+In the future, fsspec could be used as a backend to expand Pooch's download
+capabilities beyond HTTPS and FTP.
+Intake manages data catalogues (with download locations and extensive
+metadata), data download and caching (using fsspec), data loading,
+visualization, and browsing.
+It has built-in capabilities for loading data into standard containers,
+including numpy, pandas, and xarray.
+While Intake and fsspec are powerful and highly configurable,
+we argue that Pooch's advantage is in its simplicity.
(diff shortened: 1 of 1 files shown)


leouieda:
@martindurant I expanded the section discussing Intake, fsspec, and possible joint efforts. Could you please have a look and see if I missed anything (or wrote something wrong)? I hope to do Intake justice since I do think people should go to it if they need something more sophisticated than Pooch provides.

leouieda:
cc @santisoler @remram44 @hugovk @matthewturk @jrleeman @danshapero @andersy005

paper/paper.md
```diff
@@ -136,13 +136,25 @@ authentication) as well as processing functions for unpacking archives (zip or
...
+HTTPS/FTP servers, Amazon S3, Google Cloud Storage, etc).
+fsspec implements similar download and caching functionality to
+the one in Pooch, but has a wider range of download methods available.
+In the future, fsspec could be used as a backend to expand Pooch's download
+capabilities beyond HTTPS and FTP.
+Intake manages data catalogues (with download locations and extensive
+metadata), data download and caching (using fsspec), data loading,
+visualization, and browsing.
+It has built-in capabilities for loading data into standard containers,
+including numpy, pandas, and xarray.
```
hugovk:
```suggestion
including NumPy, pandas, and xarray.
```

paper/paper.md
```diff
@@ -136,13 +136,25 @@ authentication) as well as processing functions for unpacking archives (zip or
...
 To the best of the authors' awareness, the only other Python software with some
-overlapping functionality is [Intake](https://github.com/intake/intake).
-While Intake is powerful and can be used to manage large data archives,
-we argue that Pooch has a simpler setup and meets the
-specific needs of scientific software authors and individual scientists.
-For example, Pooch does not require users to change their data loading code to
-fit into a plug-in structure, instead only providing the file path for the
-user.
+overlapping functionality are [Intake](https://github.com/intake/intake) and
+[fsspec](https://github.com/intake/filesystem_spec) (which is the backend for Intake).
```
martindurant: fsspec is also standalone and used by, for example, Dask (without Intake)

paper/paper.md
```diff
@@ -136,13 +136,25 @@ authentication) as well as processing functions for unpacking archives (zip or
...
+The fsspec library provides a unified interface for defining file systems and
+opening files, regardless of where the files are located (local system,
+HTTPS/FTP servers, Amazon S3, Google Cloud Storage, etc).
+fsspec implements similar download and caching functionality to
+the one in Pooch, but has a wider range of download methods available.
+In the future, fsspec could be used as a backend to expand Pooch's download
+capabilities beyond HTTPS and FTP.
```
martindurant: Or indeed the reverse, if you wanted https://github.com/intake/filesystem_spec/issues/227

paper/paper.md
```diff
@@ -136,13 +136,25 @@ authentication) as well as processing functions for unpacking archives (zip or
...
+Intake manages data catalogues (with download locations and extensive
+metadata), data download and caching (using fsspec), data loading,
```
martindurant: (part of the confusion arises because Intake also has its own caching layer outside of fsspec, which I intend to deprecate)

paper/paper.md
hugovk: Is an extra newline needed here (and the other new text) to properly mark end of paragraph?

paper/paper.md
```diff
@@ -136,13 +136,25 @@ authentication) as well as processing functions for unpacking archives (zip or
...
+visualization, and browsing.
+It has built-in capabilities for loading data into standard containers,
+including numpy, pandas, and xarray.
+While Intake and fsspec are powerful and highly configurable,
+we argue that Pooch's advantage is in its simplicity.
+Pooch has a simpler setup and is more accessible to developers without advanced
```
martindurant: Note that fsspec has no dependencies at all (but the backends do, e.g., requests for http(s)), and is probably of a similar install size to pooch. It is also available on pypi. 

I'm not sure that "more accessible" is reasonable:

Pooch:
```
GOODBOY = pooch.create(
    # specific arguments that need to be learned
)
GOODBOY.load_registry(
    pkg_resources.resource_stream("mypackage", "registry.txt")
)
fname = GOODBOY.fetch("gravity-data.csv")
data = pandas.read_csv(fname)
```

fsspec:
```
with fsspec.open('filecache://github.com/org/package/raw/version/gravity-data.csv.gz', 
                  target_protocol:'https', cache_storage='/tmp/files/') as of:
    data = pandas.read_csv(of)
```
(not that I use compression here, handled by fsspec not pandas)
(also note that we have an experimantal [github backend](https://github.com/intake/filesystem_spec/blob/master/fsspec/implementations/github.py#L6) where you can specify branch/tab/sha)

The biggest thing that is missing is the checksum and version information. I think it would be best for everyone to implement these in fsspec with directly in the existing caching implementation, or as a backend which can read/use pooch spec and/or download mechanism.

I'd like to point out that fsspec is not exactly a niche unknown package (see download stats on [pipy](https://pypistats.org/packages/fsspec) and [conda](https://anaconda.org/search?q=fsspec)).

paper/paper.md
martindurant: PS: the `panel`-based GUI for selecting files from any backend currently lives in Intake, but should certainly be moved to fsspec.

martindurant:
PS: docs for fsspec: https://filesystem-spec.readthedocs.io

paper/paper.md
leouieda: New lines are required but I had thought of keeping this all one paragraph discussing the state of the field. I'd be happy to break as well if you think it's best.

leouieda: :+1: that is something I hadn't thought about. Left a comment on the issue with some questions/thoughts.

paper/paper.md
leouieda: :+1: thanks for catching that. Would it be more accurate to remove the `(using fsspec)` part for now?

paper/paper.md
leouieda: > I'm not sure that "more accessible" is reasonable:

The "more accessible" part should be read together with "to developers without advanced programming knowledge" but that isn't a good phrasing for what I wanted to say. I agree that the code example you posted is way nicer in fsspec than pooch. I'll try better articulate this sentence.

> The biggest thing that is missing is the checksum and version information. I think it would be best for everyone to implement these in fsspec with directly in the existing caching implementation, or as a backend which can read/use pooch spec and/or download mechanism.

OK, intake/filesystem_spec#227 makes more sense to me now :slightly_smiling_face:

paper/paper.md
leouieda: I'll clarify that it's not just a backend for Intake

leouieda:
@martindurant thanks for all the comments :+1: I made some updates to the text based on them. What do you think? 

> I think my conclusion is: we should certainly work together to produce one tool that does this thing well.

I agree. Unifying the community under a single tool is the ultimate goal. It's just not where we are at right now. It's good to have the experience of several tools first to then unify organically taking the strong points of all of them. I'll definitely give Intake and fsspec a try on my next project to get to know them better.

paper/paper.md
martindurant: You already say that Intake uses fsspec for its file handling, so probably there's no need to repeat it here - i.e., yes, can remove the parentheses. The "old" caching mechanism in Intake is not something I want to promote any more, since the fsspec one is far simpler and easier to understand.

martindurant:
> It's good to have the experience of several tools first

Agreed, and this is very much the view that we take at Anaconda. 

It is also worth making the effort, early on, to make it easier to integrate and/or merge projects when the time is right. That's why you might find it an interesting task to investigate https://github.com/intake/filesystem_spec/issues/227 (by either calling into pooch, or implementing the download with fsspec alone, using the pooch spec).
(I conspicuously failed to agree with [pyfilesystems](https://docs.pyfilesystem.org/en/latest/) on spec requirements, so I am also guilty)

leouieda:
> That's why you might find it an interesting task to investigate intake/filesystem_spec#227 (by either calling into pooch, or implementing the download with fsspec alone, using the pooch spec).

For sure! But that will probably have to wait until the summer when teaching and grading is done :slightly_frowning_face:

leouieda:
Alright, merging this in! Thanks for all the input @martindurant

PR #138, opened 2020-01-23
Blank lines

### LINKED ISSUE #89: Check for empty lines in registry files

Pooch reads in the files and their hashes from *registry files* using `Pooch.load_registry`. Right now, it fails if there is a blank line anywhere in the file (see https://github.com/fatiando/pooch/blob/master/pooch/core.py#L444).

Ideally, the code should skip any lines for which `line.strip()` is empty. While we're at it, the line number reported in the error message is off by one because `enumerate` starts from 0 so we should add one to it.

danshapero:
Fixes #89 by skipping blank lines in registry files and adds some blank lines to the test registry. Also fixes a typo in requirements-dev.txt.

**Reminders**


requirements-dev.txt
```diff
@@ -2,5 +2,5 @@
 pytest
 pytest-cov
 coverage
-sphinx=2.2.1
-sphinx_rtd_theme=0.4.3
+sphinx==2.2.1
+sphinx_rtd_theme==0.4.3
```
leouieda: This was leftover from the time when conda didn't accept `==`. Thanks for catching that!

pooch/core.py
```diff
@@ -490,20 +490,21 @@ def load_registry(self, fname):
                     line = line.decode("utf-8")
 
                 elements = line.strip().split()
-                if len(elements) > 3 or len(elements) < 2:
+                if not len(elements) in [0, 2, 3]:
```
leouieda: This is a much nicer solution than I would have come up with :)

PR #140, opened 2020-02-12
Update utils.py

zflamig:
Docs fix for os_cache

### DIFF
--- a/pooch/utils.py
+++ b/pooch/utils.py
@@ -33,7 +33,8 @@ def os_cache(project):
     r"""
     Default cache location based on the operating system.
 
-    The folder locations are defined by the ``appdirs``  package.
+    The folder locations are defined by the ``appdirs``  package
+    using the ``user_cache_directory`` function.
     Usually, the locations will be following (see the
     `appdirs documentation <https://github.com/ActiveState/appdirs>`__):
 
--- a/pooch/utils.py
+++ b/pooch/utils.py
@@ -41,7 +41,7 @@ def os_cache(project):
     * Mac: ``~/Library/Caches/<AppName>``
     * Unix: ``~/.cache/<AppName>`` or the value of the ``XDG_CACHE_HOME``
       environment variable, if defined.
-    * Windows: ``C:\Users\<username>\AppData\Local\<AppAuthor>\<AppName>\Cache``
+    * Windows: ``C:\Users\<user>\AppData\Local\<AppAuthor>\<AppName>\Cache``
 
     Parameters
     ----------
--- a/pooch/utils.py
+++ b/pooch/utils.py
@@ -38,7 +38,7 @@ def os_cache(project):
     `appdirs documentation <https://github.com/ActiveState/appdirs>`__):
 
     * Mac: ``~/Library/Application Support/<project>``
-    * Unix: ``~/.local/share/<project>`` or the value of the ``XDG_DATA_HOME``
+    * Unix: ``~/.local/share/<project>`` or the value of the ``XDG_CACHE_HOME``
       environment variable, if defined.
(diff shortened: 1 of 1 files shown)


leouieda:
Hi @zflamig thank you for submitting a PR! Could you please provide some more context for your fix? I don't know why the previous version was wrong. From the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):

>  $XDG_DATA_HOME defines the base directory relative to which user specific data files should be stored. If $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used.

zflamig:
This code calls `appdirs.user_cache_dir` which uses the `XDG_CACHE_HOME` variable: https://github.com/ActiveState/appdirs/blob/master/appdirs.py#L314

leouieda:
Oh wow, OK so the bug is actually that we're using `user_cache_dir` instead of `user_data_dir`. At this point, we have two options:

1. Fix the docs to reflect what the code actually does
2. Fix the code to do what we said it should do in the docs

Both options are not great since 1 is actually a bug and 2 breaks backwards compatibility (somewhat).  @santisoler @danshapero @matthewturk @dopplershift @hugovk I'd appreciate some feedback on this.

danshapero:
I looked up the difference between the two directories and came across [this post](https://ploum.net/207-modify-your-application-to-use-xdg-folders/). Based on that advice you could make an argument that the cache directory was the right choice to begin with. That said, if you break backwards compatibility it won't bother me.

dopplershift:
For our use in MetPy, I find the cache directory to be appropriate. Regardless, though, it's not going to break anything on our end since the whole point is to use pooch so that we don't actually care where the data end up.

matthewturk:
I think both parts of what @dopplershift says apply to our use in yt as well.

leouieda:
Thank you for pitching in everyone! OK, it seems like the more sensible thing to do is fix the docs to reflect what the code actually does:

1. Fix the paths stated in the docstring since they are for the data directory instead of cache. 
2. Mention that we use `user_cache_directory` from appdirs to make it clearer for users.

@zflamig would you mind adding these two things?

zflamig:
Updated @leouieda

pooch/utils.py
```diff
@@ -33,14 +33,15 @@ def os_cache(project):
...
     Usually, the locations will be following (see the
     `appdirs documentation <https://github.com/ActiveState/appdirs>`__):
 
-    * Mac: ``~/Library/Application Support/<project>``
-    * Unix: ``~/.local/share/<project>`` or the value of the ``XDG_DATA_HOME``
+    * Mac: ``~/Library/Caches/<AppName>``
+    * Unix: ``~/.cache/<AppName>`` or the value of the ``XDG_CACHE_HOME``
       environment variable, if defined.
-    * Windows: ``C:\Users\<user>\AppData\Roaming\<project>\<project>``
+    * Windows: ``C:\Users\<username>\AppData\Local\<AppAuthor>\<AppName>\Cache``
```
hugovk: The style check on the CI complains this line is one character too long.

Perhaps keep `<username>` as `<user>`, or add `#  noqa: W505` at the end of this docstring to disable this one particular check:
```diff
-    """
+    """  #  noqa: W505

```
Or perhaps a newline:

```diff
-    * Windows: ``C:\Users\<username>\AppData\Local\<AppAuthor>\<AppName>\Cache``
+    * Windows:
+        ``C:\Users\<username>\AppData\Local\<AppAuthor>\<AppName>\Cache``
```

leouieda:
@zflamig thank you for the contribution! It's much appreciated. Merging :rocket:

PR #144, opened 2020-02-26
Add example for fetching mutating datasets

lmartinking:
This is a bit of a hack, but you can fetch mutating datasets by using `mock.ANY` as the hash value.
Combined with daily sub directories, you can fetch data that changes potentially every day.
This PR adds a simple example of this scenario. I have discussed with @leouieda on slack.


doc/usage.rst
```diff
@@ -138,6 +138,37 @@ Alternative hashing algorithms supported by :mod:`hashlib` can be used if necess
     print(pooch.file_hash("data/c137.csv", alg="sha512"))
 
 
+Fetching Mutating Data
```
leouieda:
```suggestion
Bypassing the hash check
```

leouieda:
```suggestion
Bypassing the hash check
```

lmartinking:
Hey @leouieda I've updated the branch based on your feedback. Let me know what you think.

doc/usage.rst
```diff
@@ -138,6 +138,39 @@ Alternative hashing algorithms supported by :mod:`hashlib` can be used if necess
     print(pooch.file_hash("data/c137.csv", alg="sha512"))
 
 
+Bypassing the hash check
+------------------------
+
+Sometimes we might not know the hash of the file or it could change on the server
+periodically. In these cases, we need a way of bypassing the hash check.
+One way of doing that is with Python's ``unittest.mock`` module, and setting the
+hash value to ``ANY`` (a special object which passes all equality tests).
```
leouieda:
```suggestion
Sometimes we might not know the hash of the file or it could change on the server
periodically. In these cases, we need a way of bypassing the hash check.
One way of doing that is with Python's ``unittest.mock`` module. It defines the object
``unittest.mock.ANY`` which passes all equality tests made against it. To bypass the 
check, we can set the hash value to ``unittest.mock.ANY`` when specifying the 
``registry`` argument for :func:`pooch.create`.
```

doc/usage.rst
```diff
@@ -138,6 +138,39 @@ Alternative hashing algorithms supported by :mod:`hashlib` can be used if necess
...
+
+In this example, we are downloading a list of weather stations around Australia:
```
leouieda:
```suggestion
In this example, we want to use Pooch to download a list of weather stations around 
Australia. The file with the stations is in an FTP server and we want to store it locally 
in separate folders for each day that the code is run. The problem is that the 
``stations.zip`` file is updated on the server instead of creating a new one, so the 
hash check would fail. This is how you can solve this problem:

```

doc/usage.rst
```diff
@@ -138,6 +138,39 @@ Alternative hashing algorithms supported by :mod:`hashlib` can be used if necess
...
+
+.. code:: python
+
+    import datetime
+    from unittest import mock
```
leouieda:
```suggestion
    import unittest.mock  
```

doc/usage.rst
```diff
@@ -138,6 +138,39 @@ Alternative hashing algorithms supported by :mod:`hashlib` can be used if necess
...
+    import pooch
+
+    CURRENT_DATE = datetime.datetime.now().date()
+    
+    GOODBOY = pooch.create(
+        path=pooch.os_cache("bom_daily_stations") / CURRENT_DATE,
+        base_url="ftp://ftp.bom.gov.au/anon2/home/ncc/metadata/sitelists/",
+        # Use `mock.ANY` for the hash value if you do not care that it changes
+        registry={
+            "stations.zip": mock.ANY,
```
leouieda:
```suggestion
            "stations.zip": unittest.mock.ANY,
```

doc/usage.rst
```diff
@@ -138,6 +138,39 @@ Alternative hashing algorithms supported by :mod:`hashlib` can be used if necess
...
+    import datetime
+    from unittest import mock  
+    import pooch
+
+    CURRENT_DATE = datetime.datetime.now().date()
+    
+    GOODBOY = pooch.create(
+        path=pooch.os_cache("bom_daily_stations") / CURRENT_DATE,
+        base_url="ftp://ftp.bom.gov.au/anon2/home/ncc/metadata/sitelists/",
+        # Use `mock.ANY` for the hash value if you do not care that it changes
```
leouieda:
```suggestion
        # Use ANY for the hash value to ignore the checks
```

doc/usage.rst
```diff
@@ -138,6 +138,39 @@ Alternative hashing algorithms supported by :mod:`hashlib` can be used if necess
...
+
+In this example, we are downloading a list of weather stations around Australia:
+
+.. code:: python
+
+    import datetime
+    from unittest import mock  
+    import pooch
+
+    CURRENT_DATE = datetime.datetime.now().date()
```
leouieda:
```suggestion
    # Get the current data to store the files in separate folders
    CURRENT_DATE = datetime.datetime.now().date()
```

doc/usage.rst
```diff
@@ -138,6 +138,39 @@ Alternative hashing algorithms supported by :mod:`hashlib` can be used if necess
...
+        path=pooch.os_cache("bom_daily_stations") / CURRENT_DATE,
+        base_url="ftp://ftp.bom.gov.au/anon2/home/ncc/metadata/sitelists/",
+        # Use `mock.ANY` for the hash value if you do not care that it changes
+        registry={
+            "stations.zip": mock.ANY,
+        },
+    )
+
+Pooch will fetch the files **once** per day, even if run multiple times. If you omit ``CURRENT_DATE`` from the path,
+then Pooch will only fetch the files once, unless they are deleted from the cache.
```
leouieda:
```suggestion
Because hash check is always ``True``, Pooch will only download the file once. When running 
again at a different date, the file will be downloaded again because the local cache folder 
changed and the file is no longer present in it. If you omit ``CURRENT_DATE`` from the cache 
path, then Pooch will only fetch the files once, unless they are deleted from the cache.
```

doc/usage.rst
```diff
@@ -138,6 +138,39 @@ Alternative hashing algorithms supported by :mod:`hashlib` can be used if necess
...
+
+**NOTE**: If run over a period of time, your cache directory will increase in size, as the files are stored in daily subdirectories.
```
leouieda:
```suggestion
.. note:: 

    If run over a period of time, your cache directory will increase in size, as the 
    files are stored in daily subdirectories.
```

lmartinking:
Hi @leouieda I have committed all your suggestions and then squashed them together. I think you write very clearly, I'm afraid you have ended up pretty much writing these docs as well!

leouieda:
@lmartinking not at all! This would have never happened if you hadn't started it. I just provided some tweaks to what you already had :slightly_smiling_face: 

Thank you for this contribution! We started a new [authorship policy](AUTHORSHIP.md) recently and would like to acknowledge all contributors by adding them as authors on the Zenodo archives we make for releases. We would be more than happy to include you if you could add yourself to the `AUTHORS.md` file (full name, affiliation, ORCID). You can do that in this PR still. Of course, that is optional and I understand if you'd rather not be included.

Either way, I'm happy with this and can merge as soon as the CI checks are green. :tada:

lmartinking:
@leouieda If I contribute code, then we can think about the authors file. Otherwise, I'm happy as is. Merge if ready! :smile:

leouieda:
Alright, as you wish. Just wanted to point out that we value code and documentation contributions equally (what's the point of undocumented code?). Feel free to open a PR later adding yourself if you change your mind :slightly_smiling_face:

Thank you!

Issue #147, opened 2020-03-08
Create a function to fetch a single file


leouieda:
**Description of the desired feature**

Consider the following use case: 

> I'm a scientist/educator writing a notebook and I need to get data from this one file on the internet. I'm on Jupyter Lab, possibly on a remote server. I don't know the hash of the file and don't need to download many files, just this one. I want the file to be cached locally and not downloaded multiple times. After the first download, I want to set the hash so that I can make sure future runs of the notebook use the same version of the data. 

Right now, if you want to use Pooch for that you need to write this code:

```python
from unittest import mock

url = "https://www.ngdc.noaa.gov/mgg/ocean_age/data/2008/grids/age/age.3.6.nc.bz2"
remote_file = url.split("/")[-1]

fname = pooch.create(
    path=pooch.os_cache("meh"), 
    base_url="", 
    registry={remote_file: mock.ANY}, 
    urls={remote_file: url}
).fetch(remote_file)

print(pooch.file_hash(fname))
# Copy and paste the hash in the registry above for future runs.
```
There is a lot of repetition and feels like a hack. The same thing in fsspec would look like:

```python
# files is a not-yet-open OpenFile object. The "with" context actually opens it
with fsspec.open(url, mode="r") as f:
    # Read in the data
```

which is much more readable but it doesn't seem to cache automatically (requiring further setup) and doesn't have a hash checking mechanism (at least not that I could find in the docs).

**Proposal**

Add a function `pooch.retrieve` that would make the code above:

```python
fname = pooch.retrieve(url, hash=None)
# hash of downloaded file logged to screen and copy+pasted above for future runs
```

* Automatically cache to a `pooch.os_cache("pooch")` location without further setup.
* File names would be set to a short hash of the file content to make sure no overwriting happens. 
* The return is the file name only, keeping with the pooch way of not trying to do anything fancy. 
* Automatically printing out the hash makes it easier to download first then make sure future downloads are the same.
* Cache location and file name can be set by user if desired.
* Function takes `downloader` and `processor` arguments like `pooch.Pooch.fetch`.

This makes it easy to have all the features of Pooch for single file downloads in very little code.

The name `pooch.fetch` is nicer but it may cause confusion with `pooch.Pooch.fetch` so `retrieve` is a good alternative.

**Are you willing to help implement and maintain this feature?** Yes

santisoler:
A function like this could help to tackle #130 through a simple gallery example.

[...]

PR #149, opened 2020-03-11
Refactor cache folder creation to its own function

leouieda:
Move the code from `pooch.create` to `pooch.utils.make_local_storage`.
The function joins the path, adds the version string, checks the
environment variable, expands the user path, checks if the path is
writable (now printing different messages if we can't create the folder
vs can't create a file), and converts the string path to `pathlib.Path`.
This will make it easier to reuse this part of the code is needed.

This will pave the way for #147 

**Reminders**:

### DIFF
--- a/pooch/utils.py
+++ b/pooch/utils.py
@@ -2,6 +2,8 @@
 Misc utilities
 """
 import logging
+import os
+import tempfile
 from pathlib import Path
 import hashlib
 from urllib.parse import urlsplit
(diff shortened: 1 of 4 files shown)


pooch/utils.py
```diff
@@ -210,3 +212,58 @@ def parse_url(url):
...
+        message = (
+            "Cannot %s data cache folder '%s'. "
+            "Will not be able to download remote data files. "
+        )
+        args = [action, path]
+        if env is not None:
+            message += "Use environment variable '%s' to specify another directory."
+            args += [env]
+
+        get_logger().warning(message, *args)
```
danshapero: I realize that I'm the one who made this a warning when we switched over to using the logging API, but do you think it should become an error? If the cache directory isn't writable, the user will just hit another error right after when they try to download something. But the real cause is here and passing through with only a warning could just obscure that.

leouieda: I was thinking the same thing. If this is the case, the warning will pop up on `import` (usually when a package creates the `Pooch` instance) and an exception is only raised on `Pooch.fetch`. So if running on Jupyter/IPython there will be time to act and set the environment variable. But if running on a script then the error comes right after the warning and people might not see it.

In that case, it might be better to added the message to the error and re-raise it. What do you think?

danshapero: I was only thinking about the case of running it in a script and hadn't thought about jupyter. In any case I think that changing it from warning to error and re-raising the exception is the way to go.

leouieda: Agreed. I opened #150 to track this for a future PR instead of implementing it here to avoid changing 2 things at once. Thanks for the feedback @danshapero :+1:

PR #151, opened 2020-03-12
Refactor code for checking the file hash

leouieda:
Move code from methods to 2 dedicated functions: one to check the hash
algorithm and the other to check that the hash matches a known hash.
The hash algorithm defaults to sha256 and hash checking functions knows
to ignore the algorithm part of the hash when checking (if needed).
This simplifies the class by removing the need to add algorithms to all
hashes (retaining the legacy behaviour) and removing some duplicated
code for calculating hashes in the Pooch class.

**Reminders**:

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -9,7 +9,14 @@ import tempfile
 import ftplib
 
 import requests
-from .utils import file_hash, check_version, parse_url, get_logger, make_local_storage
+from .utils import (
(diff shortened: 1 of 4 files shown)


leouieda:
This is part of the refactoring to make #147 possible.

leouieda:
Azure is acting up but the checks seem to have passed so I'm merging this in.

PR #152, opened 2020-03-12
Add the retrive function to fetch single files

### LINKED ISSUE #147: Create a function to fetch a single file

leouieda:
Add the function `retrieve` for downloading single data files. This is much more convenient than setting up a `Pooch`. It automatically selects a unique file name and saves files to a "pooch" cache (as determined by `os_cache`). Allows passing in `known_hash=None` to bypass the hash checks. In this case, will print a message with the SHA256 hash of the downloaded file and advising users to set `known_hash` to this value.

Fixes #147 

**Reminders**:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,6 +68,25 @@ Here are Pooch's main features:
 Example
 -------
 
+For a **scientist downloading a data file** for analysis:
+
+.. code:: python
+
+    from pooch import retrieve
+
+
+    # Download the file and save it locally. Running this again will not cause
+    # a download. Pooch will check the hash (checksum) of the downloaded file
+    # against the given value to make sure it's the right file (not corrupted
+    # or outdated).
+    fname = retrieve(
+        url="https://some-data-server.org/a-data-file.nc",
+        known_hash="md5:70e2afd3fd7e336ae478b1e740a5f08e",
+    )
+
+
+For **package developers** including sample data in their projects:
+
 .. code:: python
 
     """
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -41,7 +41,9 @@ def retrieve(url, hash, path=None, fname=None, processor=None, downloader=None):
     os.makedirs(str(path), exist_ok=True)
 
 
-def _download_if_needed(url, path, fname, hash, pooch=None, processor=None, downloader=None):
+def _download_if_needed(
+    url, path, fname, hash, pooch=None, processor=None, downloader=None
+):
 
     full_path = path / fname
 
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -19,7 +19,7 @@ from .utils import (
     hash_algorithm,
     hash_matches,
     os_cache,
-    make_unique_file_name,
+    unique_file_name,
 )
 from .downloaders import HTTPDownloader, FTPDownloader
 
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -25,6 +25,175 @@ from .downloaders import choose_downloader
 
 def retrieve(url, known_hash, fname=None, path=None, processor=None, downloader=None):
     """
+    Download and cache a single file locally.
+
+    Uses HTTP or FTP by default, depending on the protocol in the given *url*.
+    Other download methods can be controlled through the *downloader* argument
+    (see below).
+
+    The file will be downloaded to a temporary location first and its hash will
+    be compared to the given *known_hash*. This is done to ensure that the
+    download happened correctly and securely. If the hash doesn't match, the
+    file will be deleted and an exception will be raised.
+
+    If the file already exists locally, its hash will be compared to
+    *known_hash*. If they are not the same, this is interpreted as the file
+    needing to be updated and it will be downloaded again.
+
+    You can bypass these checks by passing ``known_hash=None``. If this is
+    done, the SHA256 hash of the downloaded file will be logged to the screen.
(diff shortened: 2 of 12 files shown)


leouieda:
TODO:

- [ ] Name files with `hash-filename.ext` to make sure the extension is preserved and the cache is human readable.

leouieda:
This got fairly big so I'll separate it into smaller PRs to make them easier to review.

leouieda:
Alright, this is pretty much done and tests pass everywhere. I'd appreciate a review if anyone has the time.

leouieda:
> Wouldn't be nice to have an optional argument to download the file even if a version of it already exists in the cache? This could be useful for files that are regularly updated. What do you think?

Do you mean like a `force=True` argument to download the file regardless? I can see the point but I wonder if this wouldn't make it too close to what `urllib` already does. The main feature here is the hash checking and caching so I'm hesitant to include something that shouldn't be used too often.

What do you think?

pooch/core.py
```diff
@@ -15,10 +15,221 @@
...
+        prepend the hash with ``algorithm:``, for example
+        ``md5:pw9co2iun29juoh`` or ``sha1:092odwhi2ujdp2du2od2odh2wod2``. If
+        None, will NOT check the hash of the downloaded file or check if an
+        existing file needs to be updated.
+    fname : str or None
+        The name that will be used to save the file. Should NOT include the
+        full the path, just the file name (it will be appended to *path*). If
+        None, will create a unique file name using a combination of the last
+        part of the URL (assuming it's the file name) and the MD5 hash of the
+        URL. For example, ``81whdo2d2e928yd1wi22:data-file.csv``. This ensures
```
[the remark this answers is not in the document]
leouieda: Nice catch! Thanks

santisoler:
> Do you mean like a `force=True` argument to download the file regardless? I can see the point but I wonder if this wouldn't make it too close to what `urllib` already does. The main feature here is the hash checking and caching so I'm hesitant to include something that shouldn't be used too often.
> 
> What do you think?

I was thinking that would be nice to have such option in cases where we still want to use decompressors, or when we want to redownload the file without manually removing the cached file.
Yet, I don't know how often this will be used.

If you want to take time to think about this, we can merge this and add this later.

leouieda:
> If you want to take time to think about this, we can merge this and add the new argument later.

Right, you have a good point. I think I prefer to release this first and wait to see if the need arises. It's easier to add an option than to take it away later :)

README.rst
```diff
@@ -68,6 +68,25 @@ Here are Pooch's main features:
 Example
 -------
 
+For a **scientist downloading a data file** for analysis:
```
leouieda, line 71:
```suggestion
A **scientist downloading a data file** for analysis:
```

README.rst
```diff
@@ -68,6 +68,25 @@ Here are Pooch's main features:
...
+    # a download. Pooch will check the hash (checksum) of the downloaded file
+    # against the given value to make sure it's the right file (not corrupted
+    # or outdated).
+    fname = retrieve(
+        url="https://some-data-server.org/a-data-file.nc",
+        known_hash="md5:70e2afd3fd7e336ae478b1e740a5f08e",
+    )
+
+
+For **package developers** including sample data in their projects:
```
leouieda, line 88:
```suggestion
A **package developer** including sample data in their project:
```

PR #157, opened 2020-04-01
Move all hash checking to hash_matches

leouieda:
Including the exception raising controlled by new parameter `strict`.
This removes the need for re-calculating the new hash just so we can
include it in the error message. The hash calculation can be heavy
for large files so we don't want to do it twice. Also improve the error
message to include the hash algorithm and suggest what went wrong.

Working towards #152 

**Reminders**:

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -10,12 +10,10 @@ import ftplib
 
 import requests
 from .utils import (
-    file_hash,
     check_version,
     parse_url,
     get_logger,
     make_local_storage,
-    hash_algorithm,
     hash_matches,
 )
 from .downloaders import choose_downloader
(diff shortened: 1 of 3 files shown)


danshapero:
I'm assuming that the `strict = False` case will get used for #152. If you wanted to be really obsessive about branch coverage then you could add an explicit test for that but I don't think it's necessary.

leouieda:
@danshapero right, I forgot to add the extra tests to this PR. I'll add some explicit tests for this function (right now it gets tested as part of `fetch`).

Both cases are used within `fetch`: `strict=False` when checking for updates and `stritc=True` for verifying downloaded files. My plan for #152 was to allow `known_hash=None` which would always result in a `True` return value. That way bypassing the hash check would only require changing this one function.

leouieda:
Thanks for the review, by the way :slightly_smiling_face:  Much appreciated

PR #165, opened 2020-04-21
Add the SFTPDownloader class for secure FTP downloads

### LINKED ISSUE #155: SFTP Downloader

A while ago, we added an FTP Downloader to pooch. I've been wondering whether it would reasonable to add an `SFTP Downloader` as well. For SSH Authentication, I am thinking that we could use [paramiko](https://github.com/paramiko/paramiko). 

**Are you willing to help implement and maintain this feature?** Yes/No

Yes

In case there's no demand/interest for this feature, I am happy to wait until there is a need.

luke-gregor:
Used the FTPDownloader as the template. I also copied the tests for FTP. I found some toy SFTP server with an image for testing and passes. I also tested it on real data I that use. 

Fixes #155 

**Reminders**:


dopplershift:
Is this going to make pysftp a requirement to install pooch?

luke-gregor:
@dopplershift, yep. I guess one could make it an optional with a `try: import` catch.

pooch/downloaders.py
```diff
@@ -308,3 +309,118 @@ def callback(data):
...
+                        ncols=79,
+                        ascii=use_ascii,
+                        unit="B",
+                        unit_scale=True,
+                        leave=True,
+                    )
+
+                    with progress:
+                        sftp.get(
+                            parsed_url["path"], output_file, callback=progress.view_bar
```
leouieda, line 418: Could this not be a callback function instead of a class like in the FTP version?

leouieda:
@luke-gregor it would be best to make it an optional dependency. This wouldn't be so widely used so it makes more sense to avoid imposing a new dependency on everyone. You can follow what we did for tqdm (`try: .. except:` and include it in the CI builds with optional dependencies). Let me know if you need any guidance on that.

leouieda:
@luke-gregor any reason for using `pysftp` instead of `paramiko`? I'm not familiar with either so just curious about the choice.

luke-gregor:
> @luke-gregor any reason for using `pysftp` instead of `paramiko`? I'm not familiar with either so just curious about the choice.

I already had the code for `pysftp` which I originally chose based on this post: https://stackoverflow.com/questions/48434941/pysftp-vs-paramiko

doc/install.rst
```diff
@@ -33,6 +33,7 @@ Optional:
 
 * `tqdm <https://github.com/tqdm/tqdm>`__: Required to print a download progress bar
   (see :class:`pooch.HTTPDownloader`).
+* `paramiko <https://github.com/paramiko/paramiko>`__: Required for SFTP downloads
```
leouieda, line 36:
```suggestion
* `paramiko <https://github.com/paramiko/paramiko>`__: Required for SFTP downloads
  (see :class:`pooch.SFTPDownloader`).
```

doc/requirements.txt
```diff
@@ -0,0 +1 @@
+paramiko
```
leouieda, line 1: This shouldn't be required (see comments on the class below).

pooch/__init__.py
```diff
@@ -6,6 +6,11 @@
 from .downloaders import HTTPDownloader, FTPDownloader
 from .processors import Unzip, Untar, Decompress
 
+try:
+    from .downloaders import SFTPDownloader
+except ImportError:
+    pass
```
leouieda, line 12: @luke-gregor it's probably best to always have the class defined in the API but have it raise an exception if `paramiko` is missing. This avoids the need for separate documentation `requirements.txt`.

pooch/downloaders.py
```diff
@@ -49,6 +53,10 @@ def choose_downloader(url):
         "https": HTTPDownloader,
         "http": HTTPDownloader,
     }
+
+    if paramiko is not None:
+        known_downloaders.update({"sftp": SFTPDownloader})
```
leouieda, line 58: Having this will make the error message  strange because we list `SFTPDownloader` as a part of the API and it's not obvious why a user would be "Unrecognized URL protocol sftp..." if that's the case. So it can be included permanently in `known_downloaders` and we can let the class itself raise the exception.

pooch/downloaders.py
```diff
@@ -308,3 +316,113 @@ def callback(data):
             ftp.quit()
             if ispath:
                 output_file.close()
+
+
+if paramiko is not None:
```
leouieda, line 321: This check would be better in `SFTPDownloader.__ini__` so that the class always exists (which simplifies a lot of the other code in this PR).

pooch/tests/test_downloaders.py
```diff
@@ -39,6 +48,18 @@ def test_ftp_downloader():
         assert os.path.exists(outfile)
 
 
+@pytest.mark.skipif(SFTPDownloader is None, reason="requires paramiko")
```
leouieda, line 51: If implementing the changes mentioned above, this can me `skipif(paramiko is None, ...` instead.

requirements-dev.txt
```diff
@@ -4,3 +4,4 @@ pytest-cov
 coverage
 sphinx==2.2.1
 sphinx_rtd_theme==0.4.3
+paramiko
```
leouieda, line 7: This isn't required if the class always exists but may raise an exception if paramiko is missing (docs build and tests run without it)

.azure-pipelines.yml
```diff
@@ -191,7 +191,7 @@ jobs:
       Python37-optional:
         python.version: '3.7'
         PYTHON: '3.7'
-        CONDA_INSTALL_EXTRA: "codecov tqdm"
+        CONDA_INSTALL_EXTRA: "codecov  paramiko"
```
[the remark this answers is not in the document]
luke-gregor: ah, mistake. put tqdm back

pooch/__init__.py
luke-gregor: Ah, yes! Way more elegant :)

pooch/downloaders.py
luke-gregor: I inserted the following under SFTPDownloader.__init__
```suggestion
        if paramiko is None:
            raise ImportError(
                'SFTP is only supported if paramiko is installed'
            )
```

leouieda:
:wave: Hi @luke-gregor  just wanted to checkin on the status of this PR. Do you need anything from me at this point? I have a little bit more time now and could help with the development if you want. No rush, though :slightly_smiling_face:

luke-gregor:
Hey @leouieda, As far as I can see all tests are passing except `codecov`. I'll re-request a review.

pooch/downloaders.py
```diff
@@ -308,3 +314,113 @@ def callback(data):
...
+                    unit_scale=True,
+                    leave=True,
+                )
+                with progress:
+
+                    def callback(current, total):
+                        "Update the progress bar and write to output"
+                        progress.total = int(total)
+                        progress.update(int(current - progress.n))
+
```
andersy005, line 419: As a side note: these lines of code appear to be common across the `HTTPDownloader`, `FTPDownloader`. It would be nice to do some refactoring at some point in the future so as to avoid repeating ourselves.

leouieda: Agreed. This is a good target for refactoring.

leouieda:
> I am not sure how to if we can address this on our end.

Ideally, it would be great to have an FTP server that is more stable or maybe under our control but I don't think that's possible. But maybe we could use an official FTP server somewhere? Maybe one of the NOAA or UNAVCO data servers? They usually have small README files or similar that we could use as a test download. There isn't that much activity on this repo that our CIs would cause a significant impact on their servers.

leouieda:
I restarted the Azure jobs and they seem to be passing now.

pooch/downloaders.py
```diff
@@ -308,3 +314,113 @@ def callback(data):
...
+            raise ImportError("SFTP is only supported if paramiko is installed")
+
+        self.port = port
+        self.username = username
+        self.password = password
+        self.account = account
+        self.timeout = timeout
+        self.progressbar = progressbar
+        if self.progressbar and tqdm is None:
+            raise ValueError("Missing package 'tqdm' required for progress bars.")
```
leouieda, line 369: @luke-gregor the coverage report is saying that this line is never called in the tests.

pooch/downloaders.py
```diff
@@ -308,3 +314,113 @@ def callback(data):
...
+        parsed_url = parse_url(url)
+        isfile = hasattr(output_file, "write")
+
+        connection = paramiko.Transport(sock=(parsed_url["netloc"], self.port))
+
+        if isfile:
+            # paramiko requires a file name as a string, not a file object
+            # file is first closed to avoid any errors and then get name
+            output_file.close()
+            output_file = output_file.name
```
leouieda, line 395: @luke-gregor these also don't seem to be covered. Did this error happen in practice? I'm sure if Pooch ever passes an open file to a downloader.

A possible failure for this would be calling `close()` on a file inside a `with` block. Would that cause trouble when exiting the `with`? Some tests for this might be good.

If this situation never happens in normal use of Pooch, it might be best to let paramiko raise the exception and let the user deal with passing a file name instead of open file.

pooch/downloaders.py
luke-gregor: Added a new test for this - could not be "parametrized" with the existing test function as SFTPDownloader cannot be called if `paramiko` is not installed

pooch/downloaders.py
luke-gregor: Yep, sounds like a neater solution. I've removed this and added a test.

luke-gregor:
Seems that all FTP tests are failing on all the Azure tests. 

I think I need a bit of help getting codecov to run the SFTP test. It seems to be initialising the class fine, but not calling.

leouieda:
Hi @luke-gregor this was recently fixed in #192, which generates some conflicts with this branch. I'm going to update your branch and resolve conflicts. Please do a `git pull` after that. Hopefully the FTP tests should be fixed.

.gitignore
```diff
@@ -14,3 +14,4 @@ doc/api/generated
 *.egg-info
 MANIFEST
 .coverage.*
+.vscode/settings.json
```
leouieda, line 17:
```suggestion
```

@luke-gregor you might want to set this in your global `~/.gitignore_global` instead.

doc/usage.rst
```diff
@@ -482,6 +484,12 @@ explicitly to :meth:`pooch.Pooch.fetch`:
         return data
 
 
+To download files hosted on SFTP servers, the package `paramiko` needs to be
```
leouieda, line 487:
```suggestion
To download files hosted on SFTP servers, the package `paramiko <https://github.com/paramiko/paramiko>`__ needs to be 
```

pooch/downloaders.py
```diff
@@ -311,3 +317,108 @@ def callback(data):
...
             if ispath:
                 output_file.close()
+
+
+class SFTPDownloader:  # pylint: disable=too-few-public-methods
+    """
+    Download manager for fetching files over SFTP.
+
+    When called, downloads the given file URL into the specified local file
+    Uses the :mod:`paramiko` module to manage downloads.
```
leouieda, line 327:
```suggestion
    Requires :mod:`paramiko` to be installed.
```

pooch/downloaders.py
```diff
@@ -311,3 +317,108 @@ def callback(data):
             ftp.quit()
             if ispath:
                 output_file.close()
+
+
+class SFTPDownloader:  # pylint: disable=too-few-public-methods
+    """
+    Download manager for fetching files over SFTP.
+
+    When called, downloads the given file URL into the specified local file
```
leouieda, line 326:
```suggestion
    When called, downloads the given file URL into the specified local file.
```

pooch/downloaders.py
```diff
@@ -311,3 +317,108 @@ def callback(data):
...
+        self,
+        port=22,
+        username="anonymous",
+        password="",
+        account="",
+        timeout=None,
+        progressbar=False,
+    ):
+        if paramiko is None:
+            raise ImportError("SFTP is only supported if paramiko is installed")
```
leouieda, line 363:
```suggestion
            raise ValueError("Missing package 'paramiko' required for SFTP downloads.")
```

pooch/tests/test_downloaders.py
```diff
@@ -35,6 +47,38 @@ def test_ftp_downloader(ftpserver):
...
+@pytest.mark.skipif(ON_TRAVIS, reason="SFTP is not allowed on Travis CI")
+def test_sftp_downloader_fail_if_fobj():
+    "Downloader should fail when a file object rather than string is passed"
+    with TemporaryDirectory() as local_store:
+        downloader = SFTPDownloader(username="demo", password="password")
+        url = "sftp://test.rebex.net/pub/example/pocketftp.png"
+        outfile = os.path.join(local_store, "pocketftp.png")
+        outfile_obj = open(outfile, "wb")
+        with pytest.raises(TypeError):
+            downloader(url, outfile_obj, None)
```
leouieda, line 72:
```suggestion
        with open(outfile, "wb") as outfile_obj:
            with pytest.raises(TypeError):
                downloader(url, outfile_obj, None)
```

The Windows failures are coming from here. The temporary directory is trying to clean up while the file object is still open. Using a `with` statement should fix this.

requirements-dev.txt
```diff
@@ -5,3 +5,4 @@ pytest-localftpserver
...
```
leouieda, line 8:
```suggestion
```

Not necessary since we're trying to test without paramiko initially and only install it on particular jobs.

leouieda:
@luke-gregor the FTP tests are working now but there were some SFTP failures on Windows. See the comments above for a potential fix. Apart from the suggestions I have above, this looks pretty much ready to me (provided the CIs pass).

leouieda:
@luke-gregor I found the reason for the test coverage to have dropped. The problem was that we don't have a test run that has paramiko but not tqdm installed. So the test that checks if the tqdm error is raised was always skipped. To fix that, I combined both errors into one. This simplified the tests a bit as well. 

Thank you for contributing this to Pooch! Sorry the process took so long but I hope the review was enjoyable at least 🙂

PR #168, opened 2020-05-07
Added skimage to README

soupault:
We have recently switched to `pooch` as our backend for data management - https://github.com/scikit-image/scikit-image/pull/3945. Proposing to add a note on `scikit-image` as one of the libraries using `pooch`.


leouieda:
@soupault thank you for the PR! Great to see skimage adopting Pooch :slightly_smiling_face:

Issue #150, opened 2020-03-12
Raise exception instead of warning when cache isn't writable


leouieda:
**Description of the desired feature**

As pointed out by @danshapero in #149, when the cache folder isn't writable we warn the user through an error message but don't re-raise the `PermissionError` that is generated. This might be OK for interactive environments like Jupyter but it would cause an error for a script and the actual warning would be lost in the traceback.

A more sensible thing to do would be to add the message we create to the exception and re-raise it instead of logging. 

The code for this is in `pooch/core.py` in the `make_local_storage` function (after #149 is merged).

leouieda:
Actually, this is probably not a good idea. Since the cache directory is created at import time, this would cause packages to crash immediately even if the user doesn't care about sample data (see scikit-image/scikit-image#4664).

PR #173, opened 2020-05-13
Delay cache creation until download

### LINKED ISSUE #172: Delay cache folder creation to Pooch.fetch

`pooch.create` tries to ensure that the cache folder exists by creating it and trying to write to it through   `pooch.utils.make_local_storage`. This can cause problems if users call `create` to make a module global variable (since it will run at import time). If the import is run in parallel, this can lead to a race condition. #171 tried to fix this but it's not ideal since we shouldn't be writing to disk at import time anyway. A better approach would be to delay the folder creation to `Pooch.fetch` (which already does this https://github.com/fatiando/pooch/blob/master/pooch/core.py#L548). 

In this case, `make_local_storage` needs to be modified to only create the folder name (it would probably need to be renamed). This would also make #150 and comments in there unnecessary since we don't need to check for permission error if we're not writing to disk until `fetch` time. `retrieve` will also need to be modified to create the cache folder as well (like `fetch` does). We might want to wrap the folder creation so that we can check for permission error and add a message about the environment variable to the exception.

There might be unintended consequences to not checking if the cache folder is writable at import time. #22 is when we decided to do this in the first place.

See scikit-image/scikit-image#4660

leouieda:
As seen in recent issues in scikit-image, creating the cache folder in `pooch.create` can cause problems since this function is called at import time by our users. This means that importing the package in parallel can cause race conditions and crashes. To prevent that from happening, delay the creation of the cache folder until `Pooch.fetch` or `retrieve` are called. Creating the folder name was factored out of `make_local_storage` into the new function `cache_location`. Both `fetch` and `retrieve` then use `make_local_storage` to create the folder while `create` only runs `cache_location`. Since we can't check if the cache is writable when calling `create`, then it makes more sense to raise an exception with the message about using the environment variable instead of a log message (as #150 wanted).

Fixes #172 

**Reminders**:

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -309,6 +309,9 @@ def create(
     ...              registry={"data.txt": "9081wo2eb2gc0u..."})
     >>> print(pup.path.parts)  # The path is a pathlib.Path
     ('myproject', 'v0.1')
+    >>> # The local folder is only created when a dataset is first downloaded
+    >>> print(pup.path.exists())
+    False
     >>> print(pup.base_url)
     http://some.link.com/v0.1/
     >>> print(pup.registry)
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -255,8 +255,8 @@ def create(
     ``https://github.com/fatiando/pooch/raw/v0.1/data``). If the version string
     contains ``+XX.XXXXX``, it will be interpreted as a development version.
 
-    Does **not** create the local data storage folder if it doesn't exist. The
-    folder will only be created the first time a download is attempted with
+    Does **not** create the local data storage folder. The folder will only be
+    created the first time a download is attempted with
     :meth:`pooch.Pooch.fetch`. This makes it safe to use this function at the
     module level (so it's executed on ``import`` and the resulting
     :class:`~pooch.Pooch` is a global variable).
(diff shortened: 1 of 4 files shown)


leouieda:
This should fix the problems encountered in scikit-image/scikit-image#4660. I'd appreciate a review to see if I didn't do anything stupid here that would break other things :slightly_smiling_face:.

pooch/core.py
```diff
@@ -253,6 +255,12 @@ def create(
     ``https://github.com/fatiando/pooch/raw/v0.1/data``). If the version string
     contains ``+XX.XXXXX``, it will be interpreted as a development version.
 
+    Does **not** create the local data storage folder if it doesn't exist. The
```
hugovk, line 258: Can simplify a bit:
```suggestion
    Does **not** create the local data storage folder. The
```
Because it also does not create the folder if it does exist :)

leouieda: Good catch :)

leouieda:
Thanks for the comments @hmaarrfk and @hugovk! I'll merge this in and make a 1.1.1 release to get this out to the affected users.

@danshapero @andersy005 sorry for rushing this through and not waiting for your comments.

danshapero:
No worries! This seemed pretty urgent.

Issue #190, opened 2020-07-09
More stable tests for FTP downloads


leouieda:
**Description of the desired feature**

Our FTP downloader is tested on free ftp test websites. As seen recently in #189 and #165, we’ve been getting connection problems on CI from those services. It would be good to have another way of testing this that didn’t rely on these websites. 
Some possible alternatives:

1. Use an actual FTP data server run by a government organisation (NOAA or UNACO for example). But that still doesn’t solve the problem of things being out of our control on those websites as well.
2. Mock the FTP libraries and download a different data set instead. This would be more robust to failure but wouldn’t really test correct usage of the ftp libraries.

I’m open to other suggestions, particularly since neither of the above solve all our problems.

**Are you willing to help implement and maintain this feature?** Yes

leouieda:
One alternative would be to use the Debian Lenny release GPG signature: http://archive.debian.org/debian-archive/debian/dists/lenny/Release.gpg This seems to be relatively stable (last modified in 2012) and is a small file to download and test against. The Python `ftplib` docs use the debian ftp server in their doctests: https://docs.python.org/3/library/ftplib.html

leouieda:
@prisae recommended this pytest plugin: https://github.com/oz123/pytest-localftpserver It spins up a local server, which would allow us to actually test the download but without relying on an internet source. It would need to be introduced as a development dependency on `requirements-dev.txt` (still need to check if it's on conda-forge as well).

leouieda:
Ideally, pytest-localftpserver needs a conda-forge recipe since it doesn't already exist. This shouldn't be too much work since it's a pure Python package.

prisae:
If it is just testing, and not deploying, does it need to be on conda-forge?

leouieda:
Not strictly but our whole CI setup kind of assumes things are on conda-forge. At this point, it's probably easier to add this to conda-forge than to update the CI scripts :slightly_smiling_face:

leouieda:
Oh the irony, the Debian FTP server is causing `OSError: [Errno 101] Network is unreachable` using the code [here](https://docs.python.org/3/library/ftplib.html)...

PR #191, opened 2020-07-10
Fix FTP availability check when file in directory

leouieda:
If the data file is not in the base directory, the `Pooch.is_available` test was broken since we were checking for the full path in `ftp.nlst` instead of just the file name. Using the UNAVCO ftp server revealed this (ftp://data-out.unavco.org/). The change was made since the one we were using before started refusing connections on CI.

**Reminders**:


leouieda:
@andersy005 the CIs are failing because of the other tests but it seems that the availability test passes on them. This is part of a fix for #190 since I couldn't get the localftpserver tests to work on the availability checks (the local server uses a random port and haven't implemented a way to set it).

Issue #195, opened 2020-07-27
Improvements to GitHub Actions workflows


leouieda:
PR #194 introduces GitHub Actions as our continuous integration for most systems, replacing Azure Pipelines. We're still using TravisCI to do the deployment to PyPI and Github Pages. A few things that could be improved:

- [x] Deploy packages from the master branch to test.pypi.org. This let's us check the PyPI upload and catch problems before we try to do a release.
- [x] Deploy to PyPI on tag builds.
- [x] Deploy the documentation pages to gh-pages on master and tag builds. This will likely require an adaptation of our current scripts in [continuous-integration](https://github.com/fatiando/continuous-integration). We might be able to repurpuse an existing Action instead.
- [x] Add a job to test on Python 3.9-dev. Would need to use the system Python instead of miniconda and install packages with `pip`. 
- [ ] Use reviewdog to make comments inline from flake8 output (see [this comment](https://github.com/fatiando/pooch/pull/194#discussion_r460436581))

The PyPI deployments can be done with [pypi-publish](https://github.com/marketplace/actions/pypi-publish) (see [this comment](https://github.com/fatiando/pooch/pull/194#issuecomment-663577152) for details).

Of all of these, the first 3 are a priority since they would allow us to move away from TravisCI for good.

hugovk:
> 1. Deploy packages from the master branch to test.pypi.org. This let's us check the PyPI upload and catch problems before we try to do a release.
> 2. Deploy to PyPI on tag builds.

Here's an example: 

* https://github.com/hugovk/tinytext/blob/master/.github/workflows/deploy.yml

leouieda:
Deployment to PyPI was implemented in #196 and to GitHub Pages in #198.

leouieda:
Closing this since we don't really need the reviewdog bit.

PR #196, opened 2020-07-27
Add GitHub Actions workflow for publishing to PyPI

leouieda:
Use a separate workflow that doesn't run on pull requests. Builds the
archives/wheels and pushes to test PyPI on pushes to the master branch
and to PyPI proper on releases.

**Reminders**:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,8 +7,6 @@ on:
   push:
     branches:
       - master
-      # FOR TESTING ONLY! REMOVE BEFORE MERGING
-      - deploy-pypi
   release:
     types:
       - published
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,80 @@
+# Configuration for publishing archives to PyPI and documentation pages to
+# GitHub Pages using GitHub Actions
+name: deploy
+
+# Only run for pushes to the master branch and releases.
+on:
+  push:
+    branches:
+      - master
+      - deploy-pypi
+  release:
+    types:
+      - published
+
+jobs:
+
+  #############################################################################
+  # Publish built wheels and source archives to PyPI and test PyPI
+  pypi:
+    name: PyPI
+    runs-on: ubuntu-latest
+    # Only publish from the origin repository, not forks
+    if: github.repository == 'fatiando/pooch'
+    # Only run after the tests finish successfully
+    needs: test
+
+    steps:
+
+      # Checks-out your repository under $GITHUB_WORKSPACE
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # Need to fetch more than the last commit so that versioneer can
+          # create the correct version string. If the number of commits since
+          # the last release is greater than this, the version still be wrong.
+          # Increase if necessary.
+          fetch-depth: 100
+
+      # Need the tags so that versioneer can form a valid version number
+      - name: Fetch git tags
+        run: |
+          git fetch origin 'refs/tags/*:refs/tags/*'
+          git tag --list
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+
+      - name: Install requirements
+        run: python -m pip install setuptools twine wheel
+
+      - name: List installed packages
+        run: python -m pip freeze
+
+      - name: Build source and wheel distributions
+        run: |
+          python setup.py sdist bdist_wheel
+          echo ""
+          echo "Generated files:"
+          ls -lh dist/
+
+      - name: Check the archives
+        run: twine check dist/*
+
+      - name: Publish to Test PyPI
+        if: success() && github.event_name == 'pull_request'
+        uses: pypa/gh-action-pypi-publish@v1
+        with:
+          user: __token__
+          password: ${{ secrets.TEST_PYPI_TOKEN}}
+          repository_url: https://test.pypi.org/legacy/
+
+      #- name: Publish to PyPI
+        ## Only publish to PyPI when a release triggers the build
+        #if: success() && github.event.action == 'published'
+        #uses: pypa/gh-action-pypi-publish@v1
+        #with:
+          #user: __token__
+          #password: ${{ secrets.PYPI_TOKEN}}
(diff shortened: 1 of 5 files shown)


leouieda:
@hugovk @dopplershift I implemented a workflow similar to what Hugo suggested in #195. 

I set it up as a separate workflow since it's only really critical when we make a release and it would be hard to have a broken release if the master branch is passing all tests (unless the PyPI upload itself is broken, which would be caught by the test pypi upload). **Does this seem sensible?** I struggled with this decision quite a bit. 

I enabled the deploy on this branch to test PyPI for now but will remove it before merging.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+          python setup.py sdist bdist_wheel
+          echo ""
+          echo "Generated files:"
+          ls -lh dist/
+
+      - name: Check the archives
+        run: twine check dist/*
+
+      - name: Publish to Test PyPI
+        if: success()
```
dopplershift, line 69: I think `if: success()` is the default.

leouieda: 👍🏽 I thought so but just wanted to be safe

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+          password: ${{ secrets.TEST_PYPI_TOKEN}}
+          repository_url: https://test.pypi.org/legacy/
+          # Allow existing releases on test PyPI without errors.
+          # NOT TO BE USED in PyPI!
+          skip_existing: true
+
+      - name: Publish to PyPI
+        # Only publish to PyPI when a release triggers the build
+        if: success() && github.event_name == 'release'
+        uses: pypa/gh-action-pypi-publish@bce3b74dbf8cc32833ffba9d15f83425c1a736e0
```
dopplershift, line 85: Well that's paranoid...and not a bad idea.

leouieda:
TODO:

- [ ] Remove this branch from the deploy script
- [ ] Remove PyPI deploy from TravisCI config

.github/workflows/deploy.yml
leouieda: 😆  yeah that blog post linked in that comment made my skin crawl

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+
+jobs:
+
+  #############################################################################
+  # Publish built wheels and source archives to PyPI and test PyPI
+  pypi:
+    name: PyPI
+    runs-on: ubuntu-latest
+    # Only publish from the origin repository, not forks
+    if: github.repository == 'fatiando/pooch'
```
dopplershift, line 24: Technically, you *could* rely on the fact that `secrets.PYPI_TOKEN` is only available on the original repo. If someone else put their own token as a secret on their fork, it would only work if it had permission on pypi.org. Just thinking out loud...

dopplershift:
In general this makes sense, but it's also the kind of things I never get right the first time. So hopefully the pushes to test.pypi.org work...

.github/workflows/deploy.yml
leouieda: Yeah, but that would mean that the workflow runs and raises an error for most contributors when they update their forks. I get really bothered by those red crosses...

leouieda:
🤞🏽 but it seems to work for test.pypi.org so far. I guess we might have to merge this and see. But if it works on test.pypi it should work on pypi, right? Why would something like this ever break unexpectedly?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+
+    steps:
+
+      # Checks-out your repository under $GITHUB_WORKSPACE
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # Need to fetch more than the last commit so that versioneer can
+          # create the correct version string. If the number of commits since
+          # the last release is greater than this, the version still be wrong.
```
hugovk, line 34:
```suggestion
          # the last release is greater than this, the version will still be wrong.
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+          # we rely the hash to indicate to the tests that this is a local
+          # verison instead of a published version.
+          sed --in-place "s/pep440/pep440-pre/g" setup.cfg
+          python setup.py sdist bdist_wheel
+          echo ""
+          echo "Generated files:"
+          ls -lh dist/
+
+      - name: Check the archives
+        run: twine check dist/*
```
hugovk, line 66: This is part of pypa/gh-action-pypi-publish by default:

https://github.com/pypa/gh-action-pypi-publish#disabling-metadata-verification

hugovk:
> I set it up as a separate workflow since it's only really critical when we make a release and it would be hard to have a broken release if the master branch is passing all tests (unless the PyPI upload itself is broken, which would be caught by the test pypi upload). **Does this seem sensible?** I struggled with this decision quite a bit.

Yep, sounds sensible.

I have separate workflows for lint, test, and deploy. It also means if, say, there's some temporary network error which stops deploy, you can restart just the deploy workflow.

.github/workflows/deploy.yml
hugovk: Which blog post?

leouieda: This one: https://julienrenaux.fr/2019/12/20/github-actions-security-risk/

.github/workflows/deploy.yml
leouieda: Ah right. But then again, they also that we should be running this. Might just leave it since it takes a few seconds only.

leouieda:
@hugovk thanks for the comments! 

> I have separate workflows for lint, test, and deploy.

Thanks for sharing that link to your tinytext project! It really cleared up some things for me. 

> It also means if, say, there's some temporary network error which stops deploy, you can restart just the deploy workflow.

That's what I figured. Plus, I'm super paranoid with the CI on master so the release should be safe (famous last words).

leouieda:
Alright fingers crossed that this actually works 🤞🏽

leouieda:
It works! https://test.pypi.org/project/pooch/#history

PR #198, opened 2020-07-28
Push docs to GitHub Pages using Actions

leouieda:
Replaces our custom TravisCI script with GitHub Actions for pushing the built documentation to the `gh-pages` branch.

**Reminders**:


leouieda:
Download the built docs from the `test` workflow doesn't work because we can't share artefacts between workflows, only between jobs in the same workflow. So we either build the docs again or move the docs deployment to the test workflow (which is not great since there is no easy way to rerun it without running all the tests).

leouieda:
TODO: 

- [ ] Remove the current branch from `on:`
- [ ] Remove the `GHPAGE_BRANCH` variable from the checkout and push commands
- [ ] Uncomment the PyPI deploy

leouieda:
@dopplershift @hugovk this should do it for deploying to Github Pages producing the same output as we had before with Travis. @dopplershift I copied most of that script from MetPy so this might be useful if you haven't migrated this part already. 

I couldn't use the artefacts from the testing workflow because Actions doesn't support it. Instead of deploying from the test workflow I'm building the docs again here and then pushing them. It's not ideal but it's what works for now. I was trying to use https://github.com/peaceiris/actions-gh-pages but I couldn't make it reset the `latest` link. In the end, I just copied our previous shell script into the workflow. The `checkout` action actually sets the github token so you can push without extra configuration (which is nice but also alarming). 

I tested by deploying to this branch: https://github.com/fatiando/pooch/commits/gh-pages-test I made a few `dev` pushes which did amend the existing commit and "tag" push that did what it was supposed to do.

dopplershift:
I saw this and marked it as something I want to come and "learn from". Bummer that you couldn't get one of the actions to work. I might have to ponder a PR/issue to one that makes sense because I'd love not to handle that machinery (not that ours has changed in years).

What you've got here seems reasonable to me.

leouieda:
>  I might have to ponder a PR/issue to one that makes sense because I'd love not to handle that machinery (not that ours has changed in years).

I’m considering this as well. I haven’t the slightest clue how these things work though. But what would be required is the ability to run a given bash script/command inside the checked out repository. It feels like setting the latest link is a bit too niche? 

At least having the code here can serve as a guide for what we need from the Action. But as you said, this code hasn’t changed in years.

PR #202, opened 2020-07-31
Refactor the documentation into separate pages

leouieda:
Break up the usage documentation into separate pages: basic usage
(Training your Pooch with some elements removed), single file downloads,
downloaders, processors, and advanced tricks. Moved the processor and
downloader specifications from the docstring of Pooch.fetch to the
respective pages and link to them from the docstrings. Also did general
updates to the docstrings (for example, to include the new retrieve
function) and tutorials. Separated the side menu into Getting Started,
User Guide, and Reference Documentation like in all other projects.

Fixes #188 

See a rendered version of the docs here: https://www.leouieda.com/pooch-docs-refactor/

**Reminders**:

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst` and the base `__init__.py` file for the package.
- [ ] Write detailed docstrings for all functions/classes/methods. It often helps to design better code if you write the docstrings first.
- [ ] If adding new functionality, add an example to the docstring, gallery, and/or tutorials.
- [ ] Add your full name, affiliation, and [ORCID](https://orcid.org) (optional) to the `AUTHORS.md` file (if you haven't already) in case you'd like to be listed as an author on the [Zenodo](https://zenodo.org/communities/fatiando) archive of the next release.

PR #203, opened 2020-08-05
Let the user define the decompressed file name.

horta:
I don't like the decompressed file name as it is. I would like to define my own custom naming.
In particular, I want files like "foo.csv.gz" to become "foo.csv" when decompressed.

The changes I'm proposing is backward compatible and are minimal, in my opinion.

Let me know what you think and thanks a lot for this amazing package!

**Reminders**:


doc/usage.rst
```diff
@@ -375,6 +375,25 @@ keeping a decompressed copy of the file:
...
+
+    from pooch import Decompress
+
+    def fetch_compressed_file():
+        """
+        Load a large binary file that has been gzip compressed.
+        """
+        # Pass in the processor to decompress the file on download
+        fname = GOODBOY.fetch("large-binary-file.npy.gz",
+                              processor=Decompress(rename=lambda fname: fname[:-3]))
```
santisoler, line 392: Improve styling of docs and how we can remote the compress file extension, independently of the number of characters of the extension.

```suggestion
        # Pass in the processor to decompress the file on download
        fname = GOODBOY.fetch("
            large-binary-file.npy.gz",
            processor=Decompress(rename=lambda fname: os.path.splitext(fname)[0]),
        )
```

doc/usage.rst
```diff
@@ -375,6 +375,25 @@ keeping a decompressed copy of the file:
         data = numpy.load(fname)
         return data
 
+:class:`pooch.Decompress` returns ``"large-binary-file.npy.gz.decomp"`` as the
+decompressed file name by default. You can change its behaviour by passing a
+callback instead:
+
+.. code:: python
+
+    from pooch import Decompress
```
santisoler, line 384: Import `os` to use `os.path.splitext()`.

```suggestion
    import os
    from pooch import Decompress
```

pooch/processors.py
```diff
@@ -209,20 +210,28 @@ class Decompress:  # pylint: disable=too-few-public-methods
     method : str
         Name of the compression method. Can be "auto", "lzma", "xz", "bzip2",
         or "gzip".
-
+    rename : None or function
+        Defines the decompressed file name. It will be ``{fname}.decomp`` if
+        ``None``. Pass a function to define your own naming. Defaults to
+        ``None``.
```
santisoler, line 216: Improve docstring:

```suggestion
    rename : None or callable
        Function to change the name of the decompressed file. It must take 
        a single ``fname`` argument that stores the name of the compressed file.
        If ``None``, the output file is ``{fname}.decomp``.
        Defaults to ``None``.
```

pooch/tests/test_processors.py
```diff
@@ -49,6 +49,35 @@ def test_decompress(method, ext):
         check_tiny_data(fname)
 
 
+def test_decompress_rename():
+    "Check that decompression rename customization works"
+    method = "auto"
+    ext = "gz"
+    processor = Decompress(method, lambda fname: fname[:-3])
```
santisoler, line 56:
```suggestion
    processor = Decompress(method, lambda fname: os.path.splitext(fname)[0])
```

pooch/tests/test_processors.py
```diff
@@ -49,6 +49,35 @@ def test_decompress(method, ext):
         check_tiny_data(fname)
 
 
+def test_decompress_rename():
+    "Check that decompression rename customization works"
+    method = "auto"
+    ext = "gz"
```
santisoler, line 55:
```suggestion
    ext = ".gz"
```

pooch/tests/test_processors.py
```diff
@@ -49,6 +49,35 @@ def test_decompress(method, ext):
...
+    processor = Decompress(method, lambda fname: fname[:-3])
+    with TemporaryDirectory() as local_store:
+        path = Path(local_store)
+        true_path = str(path / "tiny-data.txt")
+        # Setup a pooch in a temp dir
+        pup = Pooch(path=path, base_url=BASEURL, registry=REGISTRY)
+        # Check the logs when downloading and from the processor
+        with capture_log() as log_file:
+            fname = pup.fetch("tiny-data.txt." + ext, processor=processor)
```
santisoler, line 64:
```suggestion
            fname = pup.fetch(fname + ext, processor=processor)
```

pooch/tests/test_processors.py
```diff
@@ -49,6 +49,35 @@ def test_decompress(method, ext):
...
+            lines = logs.splitlines()
+            assert len(lines) == 2
+            assert lines[0].split()[0] == "Downloading"
+            assert lines[-1].startswith("Decompressing")
+            assert method in lines[-1]
+        assert fname == true_path
+        check_tiny_data(fname)
+        # Check that processor doesn't execute when not downloading
+        with capture_log() as log_file:
+            fname = pup.fetch("tiny-data.txt." + ext, processor=processor)
```
santisoler, line 75:
```suggestion
            fname = pup.fetch(fname + ext, processor=processor)
```

pooch/tests/test_processors.py
```diff
@@ -49,6 +49,35 @@ def test_decompress(method, ext):
...
 
 
+def test_decompress_rename():
+    "Check that decompression rename customization works"
+    method = "auto"
+    ext = "gz"
+    processor = Decompress(method, lambda fname: fname[:-3])
+    with TemporaryDirectory() as local_store:
+        path = Path(local_store)
+        true_path = str(path / "tiny-data.txt")
```
santisoler, line 59:
```suggestion
        fname = "tiny-data.txt"
        true_path = str(path / fname)
```

horta: Thank you, @santisoler . Indeed the possibility of name collision must be taken into account. Appending an extension as it is makes it very unlikely to happen.

One thing to consider as well is that decompressing files like "foo.txt.gz" by standard tools will result in a "foo.txt" file. So when I first used `pooch` I was surprised that it appended an extension instead of removing the extension regarding the compression method. From the philosophy of "not surprising the user", I think the current approach is not ideal. But I do understand your concerns about collision. Maybe a warning in the parameter documentation of `rename` might be enough.

I will apply your suggestions as soon as I can. Thanks again!

leouieda (commented):
@horta thanks for implementing this! I confess that I have the same fears as @santisoler but I agree that we might be overly paranoid. There is also nothing to guarantee that the `.decomp` file doesn’t conflict as well. I’m in favor of adding this to `Decompress`. 

We should also open an issue to add checks to the processors if the file already exists and the action is “download” (meaning that the file shouldn’t be there already). This is a good indicator of a conflicting file name and we should issue a warning or exception.

A few comments on this PR:

1. The order of authors is alphabetical by last name, so you’d come right after Mathias :-)
1. Why not just pass in the desired file name instead of a callback? It would be a lot simpler (to code and test) and likely cover all desired use cases. Remember [“simple is better than complex”](https://en.wikipedia.org/wiki/Zen_of_Python). 
1. We recently completely restructured the docs (#202) so you have a conflict there. This could probably go in `doc/processors.rst` instead.

leouieda:
Hi @horta I'm just getting back from holidays and trying to catch up on OSS work. Any updates on this? Let me know if you need/want any help moving this forward.

horta:
Lets move forward. Is that looking better? I agree with the simplicity of giving a string instead of a callback.

leouieda:
@horta sorry for dropping this. This looks good! I merged in the latest changes from the master branch (basically moved the docs to `docs/processors.rst` instead of `docs/usage.rst`). I added a warning to the docs to be careful not to overwrite existing files. I also realized that we could merge the test with the previous one since they are pretty much the same code. If CIs pass, I'll happily merge this in 🙂

leouieda:
@horta thank you for adding this new feature! Hope to see you around the project in the future as wel 🙂

horta:
Well done! Thanks!

PR #204, opened 2020-08-10
Drop support for Python 3.5

### LINKED ISSUE #199: Drop support for Python 3.5
According to [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) we could be dropping support for Python 3.6 already but we still maintain tests for Python 3.5. I propose making a new release (v1.2.0) that still support 3.5+ but adding a warning this is the last release to support Python 3.5. I'm fine with retaining 3.6 a while longer and lag behind a bit on NEP29 since it's not really causing us any major headaches. 

Dropping 3.5 would allow us to use `pathlib` more confidently throughout the package without having to convert `Path` to strings all the time or reverting back to `os.path`.

@dopplershift @matthewturk @danshapero @alessiamarcolini @hmaarrfk would that be OK with you all? I don't know what your projects stance on NEP29 is but I don't expect any major headaches from this, even if your users are on 3.5 (they'll just get pooch 1.2.0 until they upgrade).

hugovk:
Fixes #199.

* Drop support for Python 3.5
* Upgrade Python syntax with https://github.com/asottile/pyupgrade with `--py36-plus`
* Update some more `.format` to f-strings

**Reminders**:

- [n/a] Add new public functions/methods/classes to `doc/api/index.rst` and the base `__init__.py` file for the package.
- [n/a] Write detailed docstrings for all functions/classes/methods. It often helps to design better code if you write the docstrings first.
- [n/a] If adding new functionality, add an example to the docstring, gallery, and/or tutorials.

### DIFF
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -79,7 +79,7 @@ download URLs) by downloading files from one or more remote servers and storing
 them in a local data cache.
 Pooch is written in pure Python and has minimal dependencies.
 It can be easily installed from the Python Package Index (PyPI) and conda-forge
-on a wide range of Python versions: 2.7 (up to Pooch 0.6.0) and from 3.6 to 3.8.
+on a wide range of Python versions: 2.7 (up to Pooch 0.6.0) and from 3.5 to 3.8.
 The integrity of downloads is verified by comparing the file's SHA-256 hash with
 the one stored in the data registry.
 This is also the mechanism used to detect if a file needs to be re-downloaded
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -79,7 +79,7 @@ download URLs) by downloading files from one or more remote servers and storing
 them in a local data cache.
 Pooch is written in pure Python and has minimal dependencies.
 It can be easily installed from the Python Package Index (PyPI) and conda-forge
-on a wide range of Python versions: 2.7 (up to Pooch 0.6.0) and from 3.5 to 3.8.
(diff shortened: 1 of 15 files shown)


paper/paper.md
```diff
@@ -79,7 +79,7 @@ download URLs) by downloading files from one or more remote servers and storing
 them in a local data cache.
 Pooch is written in pure Python and has minimal dependencies.
 It can be easily installed from the Python Package Index (PyPI) and conda-forge
-on a wide range of Python versions: 2.7 (up to Pooch 0.6.0) and from 3.5 to 3.8.
+on a wide range of Python versions: 2.7 (up to Pooch 0.6.0) and from 3.6 to 3.8.
```
hugovk, line 82: Wasn't sure whether to update this. Should it paper be kept up to date, in case a new edition is published at some point?

leouieda: Not really since we would likely need to do a big rewrite if we ever go for a second paper. There would need to be significant changes to Pooch for that to happen.

leouieda:
Thanks, @hugovk! I like the auto formatter and f-strings everywhere 🎉  I added a WIP to the title so we don’t merge this by mistake before v1.2.0 is out.

leouieda (approved):
Thanks @hugovk! I've merged in the changes from the latest master and now that 1.2.0 is out we can merge this in.

PR #214, opened 2020-11-12
Properly handle capitalized hashes

### LINKED ISSUE #193: Handle hash capitalization
Windows users will likely hit an issue when making a data manifest file as most guidance online shows finding the SHA256 of a file in power shell like this:

```
Get-FileHash myfile.bin | Format-List
```

Which then produces output like

```
Algorithm : SHA256
Hash      : 7268E3A01D57CCE828BCB7E0FF82F9048AB3A70520C58516A1145B51F7032A1D
Path      : C:\Users\myuser\myfile.bin
```

If that hash is pasted right into the manifest file pooch will try to update the file as the caps in the hash don't match strictly. I don't *believe* there would be any follow on consequences of running info read from the manifest through `.lower()` before checking for a hash match?

leouieda:
On Windows, users might sometimes get capitalized hashes from the system. To avoid false hash mismatches, convert stored and computed hashes to lowercase before doing comparisons. Convert hashes to lowercase when reading from the registry to make sure stored hashes are always lowercase. While we're at it, refactor the `hash_matches` tests to use `pytest.mark.parametrize` and allow the hash algorithm string to be upper or lowercase. 

Fixes #193 

**Reminders**:


leouieda:
@jrleeman hopefully this fixes the problem. Would you mind taking a quick look?

leouieda:
Merging this in for the next release and we can add any needed fixes if they come up. This doesn’t break anything so there should be no harm (famous last words)

PR #215, opened 2020-11-12
Add option to retry downloads if they fail

### LINKED ISSUE #205: Add option to retry downloads N times

Sometimes downloads will fail due to connection problems, etc. This can be particularly frustrating on CI services. See https://github.com/scikit-image/scikit-image/issues/4890 for an example of this happening.

It would be great to have an option for Pooch to retry a download N times if the hash check fails. We could add a configuration parameter for the `Pooch` class (like `retry_if_failed`) that can be 0 by default to retain the current behaviour and take the number N of times to retry any download. This would make it easy for projects to enable the option. The change would require:

* Adding the option to the Pooch class and the `create` function to pass it along.
* Wrapping `Pooch.fetch` in a `for` loop with a `try: ... except:` block to catch the hash mismatch error.
* Re-raise the error after N retries adding a note about it to the error message.

**Are you willing to help implement and maintain this feature?** Yes but I welcome anyone who wants to give it a shot since my time is rather limited.

leouieda:
The new `retry_if_failed` option to `create` and `Pooch` allows retrying
the download the specified number of times in case of failures due to
hash mismatches (coming from Pooch) or network issues (coming from
`requests`). This is useful for running downloads on CI that tend to
fail sporadically.

Fixes #205 

**Reminders**:


leouieda:
@jni @hmaarrfk this should resolve scikit-image/scikit-image#4890 The way it's implemented is adding the option `retry_if_fails` to `pooch.create` so you can specify the number of times to retry before giving up. You probably don't want to set this always so it could be implemented via an environment variable in skimage that sets this option to non-zero in CI.

Could you have a look and see if this would solve the problems you're having?

pooch/core.py
```diff
@@ -669,23 +693,42 @@ def download_action(path, known_hash):
...
+            if i == download_attempts - 1:
+                raise
+            retries_left = download_attempts - (i + 1)
+            get_logger().info(
+                "Failed to download '%s'. "
+                "Will attempt the download again %d time%s.",
+                str(fname.name),
+                retries_left,
+                "s" if retries_left > 1 else "",
+            )
```
hugovk, line 734: Maybe put a a small sleep here?

Network issues sometimes need a little bit of time to resolve.

---

It may not be necessary now, but another idea to consider in the future is to do something like wget, which uses linear backoff between multiple waits.

https://www.gnu.org/software/wget/manual/wget.html

leouieda: That's a good idea. From the wget man page:

> If you don’t want Wget to wait between every retrieval, but only between retries of failed downloads, you can use this option. Wget will use linear backoff, waiting 1 second after the first failure on a given file, then waiting 2 seconds after the second failure on that file, up to the maximum number of seconds you specify.
> By default, Wget will assume a value of 10 seconds. 

Those seem like good defaults (wait 1s more every time up to 10s). What do you think?

hugovk: Yep, sounds good!

leouieda: Alright, I've implemented this in  733c5fc . Seems to work fine and doesn't take too long between retries.

pooch/core.py
```diff
@@ -669,23 +698,42 @@ def download_action(path, known_hash):
...
+                raise
+            retries_left = download_attempts - (i + 1)
+            get_logger().info(
+                "Failed to download '%s'. "
+                "Will attempt the download again %d more time%s.",
+                str(fname.name),
+                retries_left,
+                "s" if retries_left > 1 else "",
+            )
+            time.sleep(i + 1 if i < max_wait else max_wait)
```
hugovk, line 739: Same as this?

```suggestion
            time.sleep(min(i + 1, max_wait))
```

hugovk (approved):

pooch/core.py
leouieda: I was thinking to myself: I know there is a better way to do this but can't think of it right now 🙂

hugovk (approved):

leouieda:
Merging this in for the 1.3.0 release. We can tweak the behavior or add missing features later if required. That would give us a chance to try this out in the wild.

jni:
Thank you @leouieda and sorry that I have been too busy to test this out, I'd intended to. Please ping again when 1.3 is released and I'll open a PR immediately on skimage to test. Thank you!!! 🙏

PR #216, opened 2020-11-13
Use pip instead of conda to speed up Actions

leouieda:
Switch to using the built-in Python and pip to setup Actions workflows
instead of conda/miniconda. This speeds up the workflows by a large
amount and is worth the trade-off since we don't have any dependencies
that require conda.

**Reminders**:

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -52,10 +52,10 @@ jobs:
             python: 3.8
             dependencies: optional
     env:
-      CONDA_REQUIREMENTS: requirements.txt
-      CONDA_REQUIREMENTS_DEV: requirements-dev.txt
-      CONDA_REQUIREMENTS_OPTIONAL: requirements-optional.txt
-      CONDA_INSTALL_EXTRA:
+      REQUIREMENTS: requirements.txt
+      REQUIREMENTS_DEV: requirements-dev.txt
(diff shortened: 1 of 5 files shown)


leouieda:
Tests now take around 1m to complete instead of 5-6minutes for Windows and Mac with conda.

hugovk:
It's possible to cache pip stuff too, something like this:

```yml
      - name: Get pip cache dir
        id: pip-cache
        run: |
          echo "::set-output name=dir::$(pip cache dir)"

      - name: Cache
        uses: actions/cache@v2
        with:
          path: ${{ steps.pip-cache.outputs.dir }}
          key:
            pip-${{ runner.os }}-${{ matrix.python }}-${{ hashFiles('requirements*.txt') }}
          restore-keys: |
            pip-${{ runner.os }}-${{ matrix.python }}-
```

Might make things a bit quicker too, and help ease the strain on PyPI just a little bit.


I also saw in some [logs](https://github.com/fatiando/pooch/pull/216/checks?check_run_id=1395552697):
```
Using legacy 'setup.py install' for pytest-localftpserver, since package 'wheel' is not installed.
Using legacy 'setup.py install' for pyftpdlib, since package 'wheel' is not installed.
```

Could also install wheel so pip'll build a wheel for those and store that in the cache for next time?

hugovk (approved):

leouieda:
Thanks for the tip on caching pip packages!

PR #217, opened 2020-11-13
Use nox to manage development automation

leouieda:
The [Nox](https://nox.thea.codes) tool allows us to specify commands
that are run inside virtual environments automatically. This lets us
run tests on multiple Python version, using conda or pip, with or
without optional dependencies, etc, locally instead of just on CI. Plus,
the configuration is written in Python (`noxfile.py`) so it's
cross-platform and easy to enhance (unlike our Makefile). The
configuration here uses requirements files in `env` to create the needed
environments for different tasks. Automation includes cleaning generated
files, producing source and wheel archives, building the docs, and
creating a sandbox conda environment to test and prototype. This would
completely replace the main `Makefile` and `doc/Makefile` (kept for
compatibility but will be removed in the future).

An added bonus is that it also simplifies our CI configuration since 
nox handles creating environments and installing packages.

**Reminders**:


leouieda:
@hugovk @remram44 @danshapero @andersy005  I'd be interested to know your opinion on this or if you have prior experience with nox or tox.

CIs are currently failing because they need updating.

hugovk:
I've used tox, it's a useful tool and quite widely used. I've not used nox; it's used by at least pip and urllib3:

https://github.com/pypa/pip
https://github.com/urllib3/urllib3

remram44:
I haven't used tox in years. It's easy enough to install different Python versions on CI, so I haven't needed it.

Looking at this PR, this doesn't look to make things simpler, it just adds another tool to the stack that contributors still have to install and deal with. It seems to me that the Makefile was good enough.

leouieda:
@remram44 that's a fair point and it's my biggest concern with using it. I kept the Makefile (which calls nox) precisely for this reason. In principle, if contributors install nox they can keep running `make format check test clean` without having to deal directly with nox. 

What is does simplify is the CI, since I could cut away the repeated code for installing dependencies and the repeated dependency specifications.

remram44:
Does it simplify that file or merely shorten it? The addition of the nox config also definitely counters that...

I feel like nox is only beneficial when it replaces the Makefile and CI config, adding it as a level of indirection means you still have to keep those places updated anyway, so I'm not sure about the benefits.

leouieda:
@remram44 is absolutely right that this adds complexity without contributing much to the development. Closing it and sticking with our current setup. Gotta stop getting overly excited about shiny new dev toys 🙂

PR #228, opened 2021-03-08
feature: add support for non-tqdm arbitrary progress bars for HTTPDownloader

neutrinoceros:
Hi ! here's a patch with minimal changes required to make the downloaders accept an arbitrary "progress bar" object from outside of pooch, making tqdm not the only way to display progress. In particular, this allows users to create custom progress bars with `rich`. The reason I'm doing this is that there are a lot of places in the project I maintain use progress bars (and the pooch downloader is one of them), and I want them to have a visual consistency.

I'm opening as a draft for now because I need to work a little more to figure out a proper way to expose the `total` computation, since it would be useful to properly define a progress from outside pooch.
I may need help to write tests (or update the existing ones) for this, but I tested it for HTTPDownloader on the project I'm maintaining (https://github.com/yt-project/yt).

Here's a screenshot of it in action
Screenshot 2021-03-08 at 18 51 41

I'll try to come up with a simple example of arbitrary progress bar class to contribute to docs later.
 
**Reminders**:

- [ ] (NA) Add new public functions/methods/classes to `doc/api/index.rst` and the base `__init__.py` file for the package.


neutrinoceros:
Apparently there's exactly one line I'm not hitting but I can figure out which one and I can't access the codecov report for some reason.

leouieda:
👋🏽 Hi @neutrinoceros thanks for the PR! This looks nice a great addition. I'm a bit strapped for time at the moment so this may take a little while to be merged. I hope that's OK.

A few quick questions: 

1. Did you make up the spec for the progress bar class or is that what is used by other libraries? It would be nice if we didn't create our own spec for this to help with interoperability.
2. Would this work if we pass in a `tqdm` instance? If so, then it would be nice to have a test for that.

neutrinoceros:
Thanks for getting back to me !
There's no rush on my side, take your time.

The specs are designed around tqdm's api to allow more flexibility in pooch with minimal change. Basically I'm going with a duck typing approach : if it looks, feels and tastes like a tqdm pbar, we can treat it as if it was, transparently and without creating much technical debt on your side.

Other libraries use slightly different method names as far as I can tell (The ones I have personally studied are tqdm and rich.progress), but it's only a matter of wrapping them with a little syntactic sugar to make them compatible.
In my own application for this, I wrapped a rich progress bar in a small custom class to make it look like tqdm and it worked :)

doc/intermediate.rst
```diff
@@ -283,3 +283,36 @@ like this:
 
     ``tqdm`` is not installed by default with Pooch. You will have to install
     it separately in order to use this feature.
+
+Alternatively, you can pass an arbitrary object that behaves like a progress
+that implements the ``update``, ``reset`, and ``close`` methods. ``update``
```
leouieda, line 288:
```suggestion
that implements the ``update``, ``reset``, and ``close`` methods. ``update``
```

doc/intermediate.rst
```diff
@@ -283,3 +283,36 @@ like this:
...
+should accept a single integer positional argument representing the current
+completion (in bytes), while ``reset`` and ``update`` do not take any argument
+beside ``self``. The object must also have a `total` attribute that can be set
```
leouieda, line 291:
```suggestion
beside ``self``. The object must also have a ``total`` attribute that can be set
```

doc/intermediate.rst
```diff
@@ -283,3 +283,36 @@ like this:
 
     ``tqdm`` is not installed by default with Pooch. You will have to install
     it separately in order to use this feature.
+
```
leouieda, line 286:
```suggestion

.. _custom-progressbar:

```

pooch/downloaders.py
```diff
@@ -84,10 +84,13 @@ class HTTPDownloader:  # pylint: disable=too-few-public-methods
 
     Parameters
     ----------
-    progressbar : bool
+    progressbar : bool or an arbitrary progress bar object
         If True, will print a progress bar of the download to standard error
         (stderr). Requires `tqdm <https://github.com/tqdm/tqdm>`__ to be
         installed.
+        Alternatively, an arbitrary object with the following methods
+        can be passed instead reset, update(int), close()
```
leouieda, line 92:
```suggestion
        Alternatively, an arbitrary progress bar object can be passed.
        See :ref:`custom-progressbar` for details.
```

pooch/downloaders.py
```diff
@@ -242,10 +248,12 @@ class FTPDownloader:  # pylint: disable=too-few-public-methods
     timeout : int
         Timeout in seconds for ftp socket operations, use None to mean no
         timeout.
-    progressbar : bool
+    progressbar : bool or an arbitrary progress bar object
         If True, will print a progress bar of the download to standard error
         (stderr). Requires `tqdm <https://github.com/tqdm/tqdm>`__ to be
         installed.
+        Alternatively, an arbitrary object with the following methods
+        can be passed instead reset, update(int), close()
```
leouieda, line 256:
```suggestion
        Alternatively, an arbitrary progress bar object can be passed.
        See :ref:`custom-progressbar` for details.
```

doc/intermediate.rst
```diff
@@ -283,3 +283,36 @@ like this:
...
+        def update(self, i):
+            self.count = i
+            self.render()
+
+        def reset(self):
+            self.count = 0
+
+        def close(self):
+            print("", file=sys.stderr)
```
leouieda, line 318: Could you add an example using this class with `HTTPDownloader`?

doc/intermediate.rst
```diff
@@ -283,3 +283,36 @@ like this:
 
     ``tqdm`` is not installed by default with Pooch. You will have to install
     it separately in order to use this feature.
+
+Alternatively, you can pass an arbitrary object that behaves like a progress
+that implements the ``update``, ``reset`, and ``close`` methods. ``update``
+should accept a single integer positional argument representing the current
+completion (in bytes), while ``reset`` and ``update`` do not take any argument
+beside ``self``. The object must also have a `total` attribute that can be set
+from outside the class.
```
leouieda, line 292:
```suggestion
from outside the class.
In other words, the custom progress bar needs to behave like a ``tqdm`` progress bar.
```

doc/intermediate.rst
leouieda: Might also be good to add a ``.. note:`` here saying that this only available for `HTTPDownloader` for now.

leouieda (changes requested):
Great stuff @neutrinoceros (and awesome handle)! Left a few suggestions, mostly for the docs. Otherwise, this seems to work like a charm 👍🏽

neutrinoceros:
Thanks for your careful review ! I'll go back to this in a few days.

doc/intermediate.rst
neutrinoceros: done !

leouieda (approved):

leouieda:
@neutrinoceros thank you for the contribution! I made a few minor edits to the docs but this is looking really good.

neutrinoceros:
Glad you like it ! Thank you for your support in getting this to the finish line. Do you intend to release it in the same go or should I except a delay ? I'm good with either way :)

leouieda:
I'm just going to push some updates to the docs and then make a release. It should be out within the next few weeks.

PR #235, opened 2021-05-05
Replace versioneer with setuptools-scm

leouieda:
Remove the versioneer code from the repository base and the package (`_version.py`). Include setuptools-scm as a setup dependency and configure it to generate the version string into a `pooch/_version.py` file whenever the package is built. We don't track this in version control and the version inference is never done at run time. This removes the need to use `pkg_resources` to get the version number and is recommended by setuptools-scm. Needed to basically invert the `MANIFEST.in` since setuptools-scm includes everything under version control in the source distribution.

Related to fatiando/maintenance#4

**Reminders**:


dopplershift (approved):
Seems reasonable to me

leouieda:
Thanks for input @santisoler and @dopplershift! I'll merge this and we'll see if the PyPI upload still works 🤞🏽

PR #241, opened 2021-06-10
Add support for figshare and Zenodo DOIs as URLs

### LINKED ISSUE #177: Download from Zenodo using a DOI instead of URL

(opening this issue per request from @leouieda )

It would be handy to have a specific downloader for data hosted on Zenodo, which could allow data to be specified in the registry by DOI rather than an absolute URL.

**Are you willing to help implement and maintain this feature?** No

But I'm willing to help test-drive it!

leouieda:
Add the `DOIDownloader` class to download files directly from figshare and Zenodo DOIs. Uses the public API of each reposiroty to get the actual HTTP download link for the given file, which is then passed to `HTTPDownloader`. Defines the new URL protocol `{repository}://{DOI}/{file name}` so that everything can happen automatically in `Pooch` and `retrieve`. 

One caveat is that we can only download 1 file from the archive and the file name must be specified in the URL. This is because our downloaders are all designed to fetch a single file and save it to a given location. Downloading multiple files would require extensive refactoring of large portions of code. This is something we can probably live with and is already pretty convenient.

The whole process was a lot easier than I expected. I basically only had to add a function that gets the download link using the API. The rest was pretty straight forward. This would serve as a good basis for other data repositories out there with DOIs and public APIs.

Fixes #177 

[Update]: Turns out it was super easy to add Zenodo as well so it's also done 🙂 

**Reminders**:

### DIFF
--- a/doc/about.rst
+++ b/doc/about.rst
@@ -20,6 +20,7 @@ Use cases
     * Make sure everyone running the code has the same version of the data
       files by verifying cryptographic hashes.
     * Multiple download protocols HTTP/FTP/SFT and basic authentication.
+    * Download from a `figshare <https://www.figshare.com>`__ DOI directly.
     * Built-in utilities to unzip/decompress files upon download
 
     **Start here:** :ref:`retrieve`
--- a/doc/about.rst
+++ b/doc/about.rst
@@ -20,7 +20,8 @@ Use cases
     * Make sure everyone running the code has the same version of the data
       files by verifying cryptographic hashes.
     * Multiple download protocols HTTP/FTP/SFT and basic authentication.
-    * Download from a `figshare <https://www.figshare.com>`__ DOI directly.
+    * Download from Digital Object Identifiers (DOIs) issued by repositories
+      like figshare and Zenodo.
     * Built-in utilities to unzip/decompress files upon download
 
     **Start here:** :ref:`retrieve`
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -38,7 +38,7 @@ Downloaders
    HTTPDownloader
    FTPDownloader
    SFTPDownloader
-   FigshareDownloader
+   DOIDownloader
 
 Processors
 ----------
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -38,6 +38,7 @@ Downloaders
    HTTPDownloader
    FTPDownloader
    SFTPDownloader
+   FigshareDownloader
 
 Processors
 ----------
--- a/doc/downloaders.rst
+++ b/doc/downloaders.rst
@@ -4,8 +4,8 @@ Downloaders: Customizing the download
 =====================================
 
 By default, :meth:`pooch.Pooch.fetch` and :meth:`pooch.retrieve` will detect
-the download protocol from the given URL (HTTP, FTP, or SFTP) and use the
-appropriate download method.
+the download protocol from the given URL (HTTP, FTP, SFTP, figshare DOI) and
+use the appropriate download method.
 Sometimes this is not enough: some servers require logins, redirections, or
 other non-standard operations.
 To get around this, use the ``downloader`` argument of
--- a/doc/downloaders.rst
+++ b/doc/downloaders.rst
@@ -4,8 +4,8 @@ Downloaders: Customizing the download
 =====================================
 
 By default, :meth:`pooch.Pooch.fetch` and :meth:`pooch.retrieve` will detect
-the download protocol from the given URL (HTTP, FTP, SFTP, figshare DOI) and
-use the appropriate download method.
+the download protocol from the given URL (HTTP, FTP, SFTP, DOI) and use the
+appropriate download method.
 Sometimes this is not enough: some servers require logins, redirections, or
 other non-standard operations.
 To get around this, use the ``downloader`` argument of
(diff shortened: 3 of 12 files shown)


santisoler (commented):
This feature is awesome! I really like it, not only because it will make things easier for me personally, but also we could invite people to upload their data to Zenodo or figshare by promoting an easy way to download it with Pooch.

The only thing I don't like of the implementation is the need to set `figshare` or `zenodo` as the _protocol_ of the url: I think it could be really confusing having them in the same set of `http` or `ftp`, while they are just strings that are being passed to tell the downloader which API should be used to fetch the desired files. I wouldn't add a separate argument for them neither.

But, what if we grab the repository out of the `doi`? As far as I know, we can use `urllib` to open any `https://doi.org/xxxx/xxxxx` url and find the original source:

```python
import urllib

url = "https://doi.org/10.6084/m9.figshare.13604360"
urllib.parse.urlparse(urllib.request.urlopen(url).geturl()).netloc
```
Returns:
```
'figshare.com'
```

This way we could avoid the need to pass `figshare` or `zenodo` to the url and also make easier to add a downloader for any other doi archives in the future: we would just need to extend the supported archives without the need to make a lot of changes on the documentation.

What do you think?

andersy005:
> This way we could avoid the need to pass figshare or zenodo to the url and also make easier to add a downloader for any other doi archives in the future: we would just need to extend the supported archives without the need to make a lot of changes on the documentation.



I concur... 

> But, what if we grab the repository out of the doi? As far as I know, we can use urllib to open any https://doi.org/xxxx/xxxxx url and find the original source:



The same applies to zenodo URLS: 

```python
In [5]: urllib.parse.urlparse(urllib.request.urlopen(url
   ...: ).geturl()).netloc
Out[5]: 'zenodo.org'

```

leouieda:
I thought about that after implementing this. The main problem is that our downloaders were designed to fetch a single file. That means that if the archive has multiple files the URL needs to specify which one we want. And a URL like `https://doi.org/10.0XXX/XXX/file.txt` doesn't really work if you put it in a browser. Downloading all contents by default would probably be ideal but it would require a lot of refactoring across the project. I'd like to avoid that if I can.

So the choices were:

1. Use `https://doi.org/DOI/file_name.txt` which won't work if  you copy it to a browser so it can be a bit confusing
2. Invent a fake protocol `repository://DOI/file_name.txt` which won't be mistaken for a real URL

I ended up going with the second but I'm not sure it's the right choice. Neither is perfect so I guess it's about choosing what we dislike the least. Or if anyone has a great workaround (not rewriting all our code :-) I'm very open to it.

leouieda:
Also very open to going with the `doi.org` option if that is what most people find best

leouieda:
Another option would be to use `doi:XXXXX/XXXX/file_name.txt` as the URL. That should be a pretty common format for anyone using DOIs and it can't be mistaken for a valid URL.

santisoler:
> Another option would be to use `doi:XXXXX/XXXX/file_name.txt` as the URL. That should be a pretty common format for anyone using DOIs and it can't be mistaken for a valid URL.

I like this option! It clearly specifies that it's not a https url but it's still agnostic of the original repository, which I think it's the ultimate goal of uploading files to a DOI instead of any other hosting service.

leouieda:
@santisoler true but the download code still needs to be specialized for each repository. For example, the code we have wouldn't support downloads from Dryad without a specific implementation. But I like this option too since the syntax is nice and clear.

@andersy005 what do you think?

andersy005:
> Invent a fake protocol repository://DOI/file_name.txt which won't be mistaken for a real URL

👍🏽. I don't think this is a big ask from a user viewpoint. I like your current implementation of this fake protocol. It's simple and maintainable :)

andersy005:
@leouieda, unless you are planning to add more work, this gets a 👍🏽 from me

leouieda:
@andersy005 @santisoler thanks for the feedback!

I'll try coding the `doi:XXX/XXX/fname.txt` format to see what that would look like. If there are any issues or if it becomes tricky to explain, I'll this as is. Otherwise, I may ping you for another quick look.

santisoler:
> @santisoler true but the download code still needs to be specialized for each repository. For example, the code we have wouldn't support downloads from Dryad without a specific implementation. But I like this option too since the syntax is nice and clear.

You're right, but if we would like to add support for downloading from Dryad, we just need to add some code to the downloader, while on the user side the code would be agnostic of the repository.
The only thing we would need to add is a error raising if the repository is not supported yet, while instructing people to ask for that repo to be supported or contribute directly to Pooch.

leouieda:
Alright, I've changed the URL format to `doi:XX.XXXX/XXXX/file_name.txt`. The code is not much different. The only problem is that Zenodo had some issue registering the DOI for our test data archive and it doesn't resolve. I added a workaround to the code that we will need to remove when that's fixed (though it doesn't affect the output if the DOI is correct). I also added a prompt for people to open issues if the repository is not supported.

What do you think?

leouieda:
Update: Zenodo resolved the issue with DOI so all is good now.

santisoler:
Looks good to me! Feel free to merge

andersy005 (approved):
This is fantastic, @leouieda! Glad to see this functionality in pooch

leouieda:
Merged 🎉  Many thanks for the reviews @santisoler and @andersy005! I'm really happy with this feature 🙂

Issue #273, opened 2021-10-28
Feature suggestion: ability to set progress bar kwarg from `fetch` or `retrieve`


GenevieveBuckley:
**Description of the desired feature:**
I think it would be useful to make the [progress bar](https://github.com/fatiando/pooch/issues/91) keyword arguments accessible from `fetch()` and/or `retrieve()`. This would allow users who don't want to do a lot of setup or configuration, to also get the benefit of progress bars (plus all the same benefits of having the right downloader chosen automatically, etc.).

It's a feature I'd like to have, so I'd like to know if there is general enthusiasm for it.

**Are you willing to help implement and maintain this feature?** 
Yes. I think it would be pretty straightforward to pass an extra kwarg through to the right functions, I can make a pull request for it.

leouieda:
👋🏽 Hi @GenevieveBuckley thanks for submitting this feature request! I agree that's pretty inconvenient to have to use `pooch.HTTPDownloader` just to get a progress bar going. It just never crossed the threshold of "I want this enough to write tests for it" for me 🙂 

Happy to get a PR for this if you're willing to implement it. As you've identified, it would be mostly adding the `progressbar=False` argument to the end of `Pooch.fetch` and `retrieve` and passing it to the downloader if one is being created. Let me know if you need any help or guidance on this. Feel free to open the PR early on in draft mode so we know you're working on this and can help if anything comes up.

PR #277, opened 2021-11-10
Allow access to progressbar kwarg from pooch.retrieve & fetch

GenevieveBuckley:
Closes https://github.com/fatiando/pooch/issues/273

This PR makes the [progress bar](https://github.com/fatiando/pooch/issues/91) keyword arguments accessible from `fetch()` and `retrieve()`. This allows users who don't want to do a lot of setup or configuration, to also get the benefit of progress bars.

**Reminders**:

- [x] ~~Add new public functions/methods/classes to `doc/api/index.rst` and the base `__init__.py` file for the package.~~ (Not applicable)

### DIFF
--- a/pooch/tests/test_downloaders.py
+++ b/pooch/tests/test_downloaders.py
@@ -54,6 +54,7 @@ ZENODOURL = pooch_test_zenodo_url()
     "sftp://test.rebex.net/pub/example/pocketftp.png",  # SFTPDownloader
 ])
 def test_progressbar_kwarg_passed(url):
+    """The progressbar keyword argument must pass through choose_downloader"""
     d = choose_downloader(url, progressbar=True)
     assert d.progressbar is True
 
--- a/pooch/tests/test_downloaders.py
+++ b/pooch/tests/test_downloaders.py
@@ -48,6 +48,7 @@ FIGSHAREURL = pooch_test_figshare_url()
 ZENODOURL = pooch_test_zenodo_url()
 
 
+@pytest.mark.skipif(tqdm is None, reason="requires tqdm")
 @pytest.mark.parametrize(
     "url",
     [
--- a/pooch/tests/test_downloaders.py
+++ b/pooch/tests/test_downloaders.py
@@ -48,6 +48,16 @@ FIGSHAREURL = pooch_test_figshare_url()
 ZENODOURL = pooch_test_zenodo_url()
 
 
+@pytest.mark.parametrize("url", [
+    BASEURL + "tiny-data.txt",  # HTTPDownloader
+    FIGSHAREURL,  # DOIDownloader
+    "sftp://test.rebex.net/pub/example/pocketftp.png",  # SFTPDownloader
+])
+def test_progressbar_kwarg_passed(url):
+    d = choose_downloader(url, progressbar=True)
+    assert d.progressbar is True
+
+
 def test_unsupported_protocol():
     "Should raise ValueError when protocol is not supported"
     with pytest.raises(ValueError):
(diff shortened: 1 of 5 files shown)


GenevieveBuckley:
Questions:
1. Is False the right default for this?

2. The pooch [progress bar documentation page](https://www.fatiando.org/pooch/latest/progressbars.html?highlight=progress) says that progress bars are only available for the HTTPDownloader. I believe this is out of date. It looks to me as though all the downloader classes have progress bar infrastructure included today. Should the docs be updated too? (This could be in a separate PR, if that makes more sense)

leouieda:
@GenevieveBuckley thank you for submitting this 🎊 

> Is False the right default for this?

Yes, that is the current default and we wouldn't want to change it. Particularly since this requires an optional dependency (tqdm).

> The pooch progress bar documentation page says that progress bars are only available for the HTTPDownloader. I believe this is out of date. It looks to me as though all the downloader classes have progress bar infrastructure included today. Should the docs be updated too? (This could be in a separate PR, if that makes more sense)

tqdm progress bars are available in the other downloaders, so the docs are outdated and need to fixed. You can fix that here since it's a minor tweak. The custom progress bars only work for `HTTPDownloader` since the implementation in #228 was restricted to that one only.

Take your time with this and feel free to ping me when you're ready or have any questions 🙂

pooch/tests/test_downloaders.py
```diff
@@ -48,6 +48,17 @@
 ZENODOURL = pooch_test_zenodo_url()
 
 
+@pytest.mark.parametrize("url", [
```
GenevieveBuckley, line 51: This test includes HTTPS, DOI, and SFTP downloaders. 
FTP is missing, because I don't see a nice way to integrate the extra setup it needs into this pytest parameterize list. You *could* manually close the FTP server if needed at the end of the pytest function, but then things get untidy if it encounters an error before that point. You'd do this if you wanted a standalone test, but I'm not sure it's worth the extra maintenance effort to include, tbh
```python
@pytest.mark.network
def test_progressbar_kwarg_passed(ftpserver):
    "Test ftp downloader"
    with data_over_ftp(ftpserver, "tiny-data.txt") as url:
    d = choose_downloader(url, progressbar=True)
    assert d.progressbar is True
```

GenevieveBuckley: Anyway, if you think this is insufficient, please leave a comment with a suggestion.

GenevieveBuckley:
> Take your time with this and feel free to ping me when you're ready or have any questions slightly_smiling_face

This is ready for review @leouieda

pooch/tests/test_downloaders.py
leouieda: @GenevieveBuckley that will be good enough, you're right that it wouldn't be worth the extra effort.

pooch/tests/test_downloaders.py
```diff
@@ -48,6 +48,21 @@
 ZENODOURL = pooch_test_zenodo_url()
 
 
+@pytest.mark.skipif(tqdm is None, reason="requires tqdm")
```
leouieda, line 51: Needed to skip this test if tqdm is not installed since it's an optional dependency (was causing some CI runs to fail).

leouieda (approved):
👋🏽 @GenevieveBuckley sorry for the delay with reviewing this. Very happy with the changes and I'm glad to merge this in! Thank you for adding yourself to the AUTHORS file as well and welcome to the contributor team 🙂 

I pushed a few changes:

1. Skip the test if tqdm is not installed
2. Fix formatting
3. Harmonized the docstrings with the downloaders classes

Nothing major but just to let you know.

Thanks again for this! 🎉

GenevieveBuckley:
:tada: Thanks @leouieda

PR #278, opened 2021-11-23
Rename "master" to "main" throughout the project

### LINKED ISSUE #276: Rename the default branch to "main"
I've been putting this off for a while now because I didn't want to deal with the headaches of renaming. But GitHub has made it quite easy now to rename the default branch so I propose we rename "master" to "main". This would require changing the setting and then updating all of the Actions and tests to make sure they use "main". Not a lot of work and anyone with forks may have to update or delete/fork again.

Any objections?

leouieda:
Update the default branch name is CI and testing. The default in the
`Pooch` class remains "master" for backwards compatibility and won't be
changed until v2.0. Updated the links to our guides as well.

Fixes #276 

**Reminders**:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -6,14 +6,14 @@
 #
 name: docs
 
-# Only build PRs, the master branch, and releases. Pushes to branches will only
+# Only build PRs, the main branch, and releases. Pushes to branches will only
 # be built when a PR is opened. This avoids duplicated buids in PRs comming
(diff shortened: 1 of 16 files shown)


santisoler (approved):
Looks great, hope we don't break any package that depends on Pooch 🤞🏼.

I've just replaced "master" for "main" on `CONTRIBUTING.md` and `MAINTENANCE.md`.
We are still defaulting to `"master"` on `pooch.create()`, which is good so we keep backwards compatibility. We should deprecate that in the future and change the default to `"main"`.

PR #291, opened 2022-01-18
Add option to disable updates

### LINKED ISSUE #290: Local-only/strict mode
The problem I'm trying to solve is that for a CI build, if the PR is changing a file but neglects to update the hash in the registry, the behavior is to silently (with default logging) re-download the file from the remote source. If there are no other tests depending on the datafile (😱) then the PR passes CI, but will fail future builds (because the committed registry and datafile are out of sync).

Would there be any interest in adding a flag to disable updating the file? My hack right now is to (on my CI builds) set `base_url` to something like `'NODOWNLOAD:'`. This is both simply just terrible and also results in errors complaining about the URL protocol rather than the true source: the bad hash.

I'd be happy to take a shot at implementing, if someone can point me to what the best approach is for this here.

dopplershift:
This gives users the option, when creating a Pooch, to disallow updating local copies of files. The intended use case for this is for development checkouts/CI systems where the local copies are canonical and any need to update likely reflects a problem with the registry (like and out of date hash).

Fixes #290.

**Reminders**:

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -263,6 +263,7 @@ def create(
     registry=None,
     urls=None,
     retry_if_failed=0,
+    allow_updates=True,
 ):
     """
     Create a :class:`~pooch.Pooch` with sensible defaults to fetch data files.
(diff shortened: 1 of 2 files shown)


leouieda (approved):
👍🏽 looks great @dopplershift! Merging this in. I'll make a new release as soon as I can.

PR #308, opened 2022-05-17
CITATION.cff from JOSS article

rowanc1:
Recently went through @leouieda's open-science tutorial/presentation [here](https://www.leouieda.com/2022-05-06-spin-open-science/#/25) and noticed no CFF files mentioned, this will add the following in the github UI:

image

See:
* https://citation-file-format.github.io
* https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

I have taken the authors from the JOSS doi article.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,63 @@
+cff-version: 1.2.0
+title: 'Pooch: A friend to fetch your data files'
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+authors:
+  - given-names: Leonardo
+    family-names: Uieda
+    affiliation: University of Liverpool
+    orcid: 'https://orcid.org/0000-0001-6123-9515'
+  - given-names: Santiago
+    family-names: Rubén Soler
+    affiliation: Universidad Nacional de San Juan
+    orcid: 'https://orcid.org/0000-0001-9202-5317'
+  - given-names: Rémi
+    family-names: Rampin
+    affiliation: New York University
+    orcid: 'https://orcid.org/0000-0002-0524-2282'
+  - given-names: Hugo
+    name-particle: van
+    family-names: Kemenade
+    orcid: 'https://orcid.org/0000-0001-5715-8632'
+  - given-names: Matthew
+    family-names: Turk
+    affiliation: School of Information Sciences
+    orcid: 'https://orcid.org/0000-0002-5294-0198'
+  - given-names: Daniel
+    family-names: Shapero
+    affiliation: University of Washington
+    orcid: 'https://orcid.org/0000-0002-3651-0649'
+  - given-names: Anderson
+    family-names: Banihirwe
+    affiliation: National Center for Atmospheric Research
+    orcid: 'https://orcid.org/0000-0001-6583-571X'
+  - given-names: John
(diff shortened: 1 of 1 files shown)


rowanc1:
I actually think the existing CITATION.rst is more useful, and I am actually a bit disappointed that the github UI does not show the DOI in the bibtex or APA citation, which basically means it is actually worse from a FAIR perspective (they opt to point to github instead, which I am sure is a battle that someone lost).

You can test it here:
https://github.com/rowanc1/pooch/tree/patch-2

Maybe some feedback for the github folks?! (Feel free to close this, as I think the UI is actually worse/misleading if there is an existing DOI.)

leouieda:
@rowanc1 thanks for taking the time to do this! The mention was missing from that talk because I completely forgot about it 😨 

To be honest, every time I thought about adding one to our repos I got put off by the huge amount of reading I would have to do to figure out the format. There are so many options and a lot of the examples are for Zenodo archives when I actually want a paper to be cited. I was actually waiting for either an automated tool that generates the file given a DOI or someone like you to stop by and add the file 😉 

The citation from GitHub is completely wrong actually:

> Uieda, L., Rubén Soler, S., Rampin, R., van Kemenade, H., Turk, M., Shapero, D., Banihirwe, A., & Leeman, J. (2020). Pooch: A friend to fetch your data files [Computer software]. https://github.com/fatiando/pooch

There is no journal name, issue, etc. I don't know enough about the CFF but maybe that's from the file saying that the citation is software instead of a paper: https://github.com/rowanc1/pooch/blob/patch-2/CITATION.cff#L6 ?

If it's really GitHub not doing a proper job, then this is probably the place for feedback: https://github.com/github/feedback/discussions/

florian-wagner:
Yes, @leouieda, you are correct. If the citation type is `article`, it works as expected. So the APA and BibTeX entries for the pyGIMLi paper look correct (except for the empty brackets after the volume, since the article has no issue number...). I had to specify it as an additional reference (which you can also make the preferred citation, so that it's picked up by the GitHub UI).

https://github.com/gimli-org/gimli/blob/dev/CITATION.cff

leouieda:
Thanks @florian-wagner, I put the information under a `preferred-citation` of type `article`. Seems to make worked well:

> Uieda, L., Rubén Soler, S., Rampin, R., van Kemenade, H., Turk, M., Shapero, D., Banihirwe, A., & Leeman, J. (2020). Pooch: A friend to fetch your data files. Journal of Open Source Software, 5(45), 1943. https://doi.org/10.21105/joss.01943

```bibtex
@article{Uieda_Pooch_A_friend_2020,
  author = {Uieda, Leonardo and Rubén Soler, Santiago and Rampin, Rémi and van Kemenade, Hugo and Turk, Matthew and Shapero, Daniel and Banihirwe, Anderson and Leeman, John},
  doi = {10.21105/joss.01943},
  journal = {Journal of Open Source Software},
  number = {45},
  pages = {1943},
  title = {{Pooch: A friend to fetch your data files}},
  volume = {5},
  year = {2020}
}
```

leouieda:
OK, merging this in! We can figure out what to do with the CITATION.rst file later.

PR #316, opened 2022-07-18
Rewrite archive extraction with members argument

### LINKED ISSUE #314: Repeated fetching with differing Untar processors yields wrong results

The use case is to download an archive, unpack it and get access to individual files within that archive through the `members` parameter of `pooch.Untar`. When doing so, the `members` argument of the first call is cached for subsequent calls, which prohibits a usage pattern that looks otherwise valid. `pooch.Unzip` is likely to have the same issue.

```python
import pooch
P = pooch.create(
    path=pooch.os_cache("test"),
    base_url="https://github.com/ssciwr/afwizard-test-data/releases/download/2022-06-09/",
    registry={
        "data.tar.gz": "sha256:fae90a3cf758e2346b81fa0e3b005f2914d059ca182202a1de8f627b1ec0c160"
    }
)
files1 = P.fetch("data.tar.gz", processor=pooch.Untar(members=["testfilter1.json"]))
files2 = P.fetch("data.tar.gz", processor=pooch.Untar(members=["testfilter2.json"]))
assert files1[0] != files2[0]
```

**Full error message**

Above assertion fails.

dokempf:
So far, the logic whether extraction needs to be run depended on the target
directory existing on the filesystem. This made it impossible to extract multiple
times from the same archive with differing members. The commit changes the
behaviour to checking whether all required files are already present.

In the process, I had to tweak the logic that compares paths and the specified
members. The implemented version now works across a variety of user inputs
(e.g. with/without leading `./` and on all platforms, regardless of path separators).

This fixes #314

### DIFF
--- a/pooch/processors.py
+++ b/pooch/processors.py
@@ -104,7 +104,9 @@ class ExtractorProcessor:  # pylint: disable=too-few-public-methods
         fnames = []
         for path, _, files in os.walk(self.extract_dir):
             for filename in files:
-                relpath = os.path.join(os.path.relpath(path, self.extract_dir), filename)
+                relpath = os.path.join(
+                    os.path.relpath(path, self.extract_dir), filename
+                )
                 relpath = os.path.relpath(relpath, ".")
                 if self.members is None or any(
                     relpath.startswith(m) for m in self.members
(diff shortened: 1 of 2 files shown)


leouieda (approved):
@dokempf looks good to me! Thank you for adding the support for the leading `./` as well 🥇 

Merging in the main branch and merging when tests pass.

PR #318, opened 2022-07-25
Dataverse doi resolution

### LINKED ISSUE #317: Add support for DOI resolution for DataVerse repositories
**Description of the desired feature:**

I would like to add DOI handling support for data repositories based on [DataVerse](https://github.com/IQSS/dataverse). My main motivation is that my institution offers a data archiving service based on DataVerse. I have done a previous [first attempt](https://github.com/ssciwr/dataverse_pooch) which was based on generating registry and urls dictionaries in a wrapper around `pooch.create`. I have come to realize however that it would be a better fit to include this into `pooch` itself alongside the other DOI resolvers (plus I read that you are open to such contributions).

There is one technical difference between a DataVerse DOI converter compared to the existing ones: It cannot decide based on the URL whether it is a DataVerse repository, as institutions will typically run their own DataVerse instances (e.g. [this at Heidelberg University](https://heidata.uni-heidelberg.de/)). To account for this difference I think a refactoring of the existing DOI converter dispatch is required. I propose the following:

*Current implementation*: Converters are stored as a [mapping of base URL to a function](https://github.com/fatiando/pooch/blob/main/pooch/downloaders.py#L558:L561) -> the base URL is determined -> the correct function is called. Converter functions return a URL.

*Proposed implementation*: Converters are stored as a list -> Converters are called one after another until one of them returned a URL, they return `None` otherwise. If no converter works, an error is thrown (similar to what is currently done). This can be seen as an implementation of the [Chain of Responsibility Design Pattern](https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern).

**Are you willing to help implement and maintain this feature?** 

I will be able to contribute the implementation of this feature. Given that I have a natural interest in the feature for supporting researchers at my institution, I will also be around for helping with maintenance.

dokempf:
**Relevant issues/PRs:** This fixes #317.

This adds support for DOI conversion from DataVerse repositories. In order to do so, it contains two refactorings which are in isolated commits:

* DOI resolution is now a chain of responsibility (see #317)
* `parse_url`'s DOI branch so far assumed exactly one `/` being present in a DOI, which is not true for Dataverse DOI's

The testing code accesses a published dataset hosted at my institution: https://heidata.uni-heidelberg.de/dataset.xhtml?persistentId=doi:10.11588/data/TKCFEF


dokempf:
Test failure is unrelated (`test_check_availability_on_ftp` failing with `TimeoutError`). This implements #317 and the necessary refactoring bits required for/discussed in #319. @leouieda  How about we try to review/merge this and I implement #319 on top of it?

leouieda:
@dokempf I'll have a look at this ASAP. In the meantime, I managed to fix the broken test in #322 so let's wait for that to be merged.

leouieda:
Merging in the main branch here to fix the broken tests.

dokempf:
I excluded the two lines from the `DataRepository` base class from coverage testing. I think this is a sensible thing to do and should make CI pass.

leouieda (commented):
👋🏾 @dokempf sorry for the long delay on this. Just got around to looking at the code yesterday and this morning. 

First of all, big thank you for putting in the effort to implement this! I like where this is going but there is something in the `*Repository` classes that seems off to me but I can't quite articulate why. Not that it's wrong or doesn't work in any way. I think I just need to play with the code a bit myself.

I'll merge this in so you can work on #319. In the meantime, I'll have a think about the repository classes and will follow up with a PR (or not 🙂).

Excellent work! 🚀

dokempf:
@leouieda Thanks for merging. I will continue with #318 and happily take any feedback on the general approach with the `*Repository` design.

PR #320, opened 2022-08-12
DOC: Add SciPy to Projects using Pooch

AnirudhDagar:
Hi, we recently ([merged](https://github.com/scipy/scipy/commit/d58e74655e1129d5d18f034b5a0219dba90368d3) just today) added a `datasets` submodule in SciPy which will be released in [SciPy 1.10](https://github.com/scipy/scipy/wiki/%2ARelease-Note-Entries-for-SciPy-1.10.0#scipydatasets-introduction), with Pooch as an optional dependency. I thought it would be nice to include that in Pooch's readme doc.

Ps. Thanks for this great, easy-to-use tool. Quite useful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ def fetch_gravity_data():
 
 ## Projects using Pooch
 
+* [SciPy](https://github.com/scipy/scipy)
 * [scikit-image](https://github.com/scikit-image/scikit-image)
 * [MetPy](https://github.com/Unidata/MetPy)
 * [icepack](https://github.com/icepack/icepack)


leouieda:
@AnirudhDagar that is great to hear! Now we also have to worry about breaking scipy when we change things. No pressure 😰

Let us know how it's working out and if there is anything we can do on the Pooch side to make things easier.

AnirudhDagar:
> Now we also have to worry about breaking scipy when we change things.

Haha yeah, always a good thing, but that also means maybe more contributions/bug fixes from the scipy community :P 

We are still discussing a few design decisions and mechanics of the `scipy.datasets` submodule, specifically how the import for `datasets` will work since `pooch` is an optional dep for SciPy and only affects the datasets submodule.

leouieda:
> but that also means maybe more contributions/bug fixes from the scipy community :P

That would be great!

Would you mind pinging me in some of those conversations? It's good to know the various use cases so we have them in mind when planning changes.

AnirudhDagar:
Sure, I'll keep that in mind. If there is something `pooch` specific, I'll ping you, would be nice to have you in the discussion. 🙌🏼

PR #321, opened 2022-08-30
Add napari to the list of Projects using Pooch

kephale:
This adds napari to the list of Projects using Pooch because napari does use Pooch!

**Relevant issues/PRs:**

N/A


leouieda:
Thanks @kephale!

PR #323, opened 2022-09-09
Add mne-python to the list of Projects using Pooch

myd7349:
This patch adds [mne-python](https://github.com/mne-tools/mne-python) to the list of Projects using Pooch.

https://github.com/mne-tools/mne-python#dependencies

https://github.com/mne-tools/mne-python/blob/v1.0.3/mne/datasets/_fetch.py#L229

**Relevant issues/PRs:**

N/A.


leouieda:
Thanks for adding @myd7349! It's great to know of more use cases for Pooch 🙂

PR #328, opened 2022-10-16
Speed up import type (0.05 s to 0.02 s) by lazy loading requests

hmaarrfk:
I'm proposing lazy loading requests since it adds a bit of import of.

Usually, I think that pooch is used as an import addon but not as a core feature. As such it is a little silly to expand the import time by alot. While small, it is not insignificant.

Thank you for considering

Current
image

Proposed
image

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -755,6 +755,7 @@ def stream_download(url, fname, known_hash, downloader, pooch=None, retry_if_fai
     will retry the download the specified number of times in case the failure
     was due to a network error.
     """
+    # Lazy import requests to speed up import time
     import requests.exceptions  # pylint: disable=C0415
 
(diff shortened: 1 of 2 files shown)


leouieda:
@hmaarrfk sorry for the long delay on this. I've been away from OSS work for a while and am only slowly returning now. We'll get to this PR as soon as possible. Thanks for going thought the trouble and being patient 🙂

leouieda:
@hmaarrfk will merge this ASAP. We're having a failure on the Mac builds for Python 3.7 because the GitHub `setup-python` action is trying to compile Python and failing to link to libbz2...

hmaarrfk:
No issues.  Take your time.

leouieda:
Thanks @hmaarrfk! Finally got some time to sort this out. Thank you your help!

Issue #319, opened 2022-07-25
Populate registry via data repository API


dokempf:
**Description of the desired feature:**

When implementing #318, I realized that the DOI support in pooch currently does not allow to automatically populate the pooch's registry using the API, although the information (incl. checksums) is readily available (checked DataVerse and Figshare, Zenodo should have the same). I would like to add this at least for DataVerse, as only with this feature, the built-in pooch solution is superior to my [prior implementation](https://github.com/ssciwr/dataverse_pooch). It would make sense however to do this in a unified way across different data repositories.

*Proposal*: Add a function `registry_from_doi` to that performs a dispatch based on the data repository type to implementation functions that map a DOI to a registry dictionary. Sample code:

```python
MYPOOCH = pooch.create(base_url="doi:..", ...)
pooch.registry_from_doi(MYPOOCH)
```
There are many ways to formulate this interface. I would prefer one that does not need redundant specification of the DOI (so it has the pooch object). Alternatively, a special marker object could be passed to `pooch.create`'s `registry=` argument.

The above proposal could be combined with a refactoring that abstracts data repositories into objects with a well-defined interface, instead of juggling around with functions. This is a matter of personal taste though.

**Are you willing to help implement and maintain this feature?** 

Yes, I can implement it and will be around for helping with maintenance.

leouieda:
@dokempf that's a good idea. And it would make downloading from DOIs a bit easier since getting the file name and hash by hand is a bit tedious.

How about:

* Keep the doi specification in the `base_url`
* Add the method `Pooch.load_registry_from_doi()` that takes no arguments and assumes that the `base_url` is a DOI (should raise an error if it's not).

This way it's consistent with the current registry population from a file.

> The above proposal could be combined with a refactoring that abstracts data repositories into objects with a well-defined interface, instead of juggling around with functions. This is a matter of personal taste though.

It would be best to do the refactoring in a follow up to avoid making this PR too large and hard to review. But the DOI interface could probably be improved, as you've pointed out. One thing is that I would be a bit hesitant to have something coming close to a full blow API wrapper for each repository (which should probably be a separate package that we then use).

Thanks for all the work you've been doing on Pooch!

dokempf:
Thanks for your comments @leouieda - I agree with everything. I am on parental leave until September 5th, so I will only be able to work on this after that date. I guess we can move on with #318 in the meantime and do the refactoring and feature implementation from this issue after that has been merged.

leouieda:
[...]

PR #337, opened 2023-01-30
Support DOIs with slashes in the filename

### LINKED ISSUE #336: error downloading file from sub-directory in zenodo repository

Thanks very much for this very useful project!

I'm trying to download a file from the zenodo repository at `doi:10.5281/zenodo.7352326` ([link](https://doi.org/10.5281/zenodo.7352326)). The file name is `GreatApeDictionary-v1.1.zip` and it seems to be in a `Wild-Minds` sub-directory in the repository. I receive an error that the file is "not found in data archive" (see below for full error message).

Looking in the `pooch/downloaders.py` code (`v1.6.0`), it seems that the `files` information that is returned by the zenodo API has the filename as `Wild-Minds/GreatApeDictionary-v1.1.zip` - but the file name that is passed to the downloader is the last item of a `split("/")` operation and so loses the `Wild-Milds` part. If I replace the [line](https://github.com/fatiando/pooch/blob/6cf0208b4bd6e1135fc96749059b1f208f28dd0c/pooch/downloaders.py#L575) `file_name=parsed_url["path"].split("/")[-1],` with `file_name="/".join(parsed_url["path"].split("/")[1:]),` then it seems to work fine.

I noticed that the code has changed in the current git version, so I tried that also (`v1.6.0.post16+gf20fe3a`). I get a different error: `ValueError: Archive with doi:10.5281/zenodo.7347607/Wild-Minds not found (see https://doi.org/10.5281/zenodo.7347607/Wild-Minds). Is the DOI correct?` - so the sub-directory seems to be being included as part of the DOI.

Apologies if I have just missed something - thanks for any assistance.

```python
raw_data_path = pooch.retrieve(
    url="doi:10.5281/zenodo.7347607/Wild-Minds/GreatApeDictionary-v1.1.zip",
    known_hash="md5:8c40d1bce25619548f4daa16d63f36a3",
)
```

**Full error message**

```
Downloading data from 'doi:10.5281/zenodo.7347607/Wild-Minds/GreatApeDictionary-v1.1.zip' to file '/home/damien/.cache/pooch/66afcc98f960ee876a64b8ab89dbec3e-GreatApeDictionary-v1.1.zip'.
  File "/home/damien/venv_study/ape_gestures/code/ape_gestures/ape_gestures/data.py", line 15, in <module>
    get_raw_data_path()
  File "/home/damien/venv_study/ape_gestures/code/ape_gestures/ape_gestures/data.py", line 6, in get_raw_data_path
    raw_data_path = pooch.retrieve(
  File "/home/damien/venv/ape_gestures/lib/python3.9/site-packages/pooch/core.py", line 240, in retrieve
    stream_download(url, full_path, known_hash, downloader, pooch=None)
  File "/home/damien/venv/ape_gestures/lib/python3.9/site-packages/pooch/core.py", line 772, in stream_download
    downloader(url, tmp, pooch)
  File "/home/damien/venv/ape_gestures/lib/python3.9/site-packages/pooch/downloaders.py", line 573, in __call__
    download_url = converters[repository](
  File "/home/damien/venv/ape_gestures/lib/python3.9/site-packages/pooch/downloaders.py", line 633, in zenodo_download_url
    raise ValueError(
ValueError: File 'GreatApeDictionary-v1.1.zip' not found in data archive https://zenodo.org/record/7352326 (doi:10.5281/zenodo.7347607).
```

**System information**

* Operating system: Linux
* Python installation (Anaconda, system, ETS): System
* Version of Python: 3.9.16
* Version of this package: 1.6.0

santisoler:
Allow DOI downloaders to work with files uploaded to Zenodo that have a forward slash "/" in their name. These special cases are generated by the Zenodo-GitHub integration service, where the first portion of the filename matches the GitHub handle of the repository owned and the last portion matches the repository name. When creating the url to these repositories, the forward slashes that are part of the filename should be led by a backslash. For example: `url = "doi:10.123/zenodo.123123/my-file\/has-a-slash.zip"`.

**Relevant issues/PRs:**

Fixes #336

**Todo**

In case this patch works, we need to:

- [x] Try to add a file with a slash to our test repos in:
    - [x] Zenodo: [10.5281/zenodo.7632643](https://doi.org/10.5281/zenodo.7632643)
        - It was created through the Zenodo-GitHub integration service. By default the zip file includes the user handle, a forward slash and the repo name.
- [x] Update tests to see if we can download them
- [ ] Improve docs to instruct on how to use this feature
    - Extend the docstring of `DOIDownloader`
    - Extend the docstring of `retrieve`
    - Add an example in the user guide

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -13,6 +13,7 @@ order by last name) and are considered "The Pooch Developers":
 * [Dominic Kempf](https://github.com/dokempf) - Scientific Software Center, Heidelberg University, Germany (ORCID: [0000-0002-6140-2332](https://www.orcid.org/0000-0002-6140-2332))
 * [Kacper Kowalik](https://github.com/Xarthisius) - National Center for Supercomputing Applications, University of Illinois at Urbana-Champaign, USA (ORCID: [0000-0003-1709-3744](https://www.orcid.org/0000-0003-1709-3744))
 * [John Leeman](https://github.com/jrleeman)
+* [Björn Ludwig](https://github.com/BjoernLudwigPTB) - Physikalisch-Technische Bundesanstalt, Germany (ORCID: [0000-0002-5910-9137](https://www.orcid.org/0000-0002-5910-9137))
 * [Daniel McCloy](https://github.com/drammock) - University of Washington, USA (ORCID: [0000-0002-7572-3241](https://orcid.org/0000-0002-7572-3241))
 * [Rémi Rampin](https://github.com/remram44) - New York University, USA (ORCID: [0000-0002-0524-2282](https://www.orcid.org/0000-0002-0524-2282))
 * [Clément Robert](https://github.com/neutrinoceros) - Institut de Planétologie et d'Astrophysique de Grenoble, France (ORCID: [0000-0001-8629-7068](https://orcid.org/0000-0001-8629-7068))
--- a/pooch/downloaders.py
+++ b/pooch/downloaders.py
@@ -623,7 +623,7 @@ class DOIDownloader:  # pylint: disable=too-few-public-methods
             )
 
         # Resolve the URL
-        file_name = parsed_url["path"].split("/")[-1]
+        file_name = parsed_url["path"][1:]  # remove the leading slash in the path
         download_url = data_repository.download_url(file_name)
 
         # Instantiate the downloader object
--- a/pooch/tests/test_downloaders.py
+++ b/pooch/tests/test_downloaders.py
@@ -34,6 +34,7 @@ from ..downloaders import (
     DataverseRepository,
     doi_to_url,
 )
+from ..processors import Unzip
 from .utils import (
     pooch_test_url,
     check_large_data,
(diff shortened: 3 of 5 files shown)


santisoler:
@djmannion

With the patch in this PR you should be able to download your zip file that have a slash in its name through:

```python
raw_data_path = pooch.retrieve(
    url=r"doi:10.5281/zenodo.7347607/Wild-Minds\/GreatApeDictionary-v1.1.zip",
    known_hash="md5:8c40d1bce25619548f4daa16d63f36a3",
)
```

I decided to use a leading backslash before the offending slash in the filename to separate it from the other slashes that are part of the url. In order to use a backslash in a Python string without meaning to use it as a escape character we need to use r-strings with `r""`.

It works locally for me, does it for you?

djmannion:
Yes, that works for me now - thanks!

santisoler:
@leouieda would you like to take a look at this approach for solving this issue? I think the backslash is a good compromise, but would love to hear your input.

It worth noting that I haven't changed the filename of the cached file. Currently it will contain only the portion of the original filename that follows the last slash (e.g. if the filename is `my/file.zip`, the cached file will be `<HASH>-file.zip`).
I think this is ok as it is, considering that when downloading the file from Zenodo webapp it uses this portion to name the file. But maybe we would like to add a test for this behaviour.

santisoler:
Well, this is trickier to solve than I expected. This patch works well with `pooch.retrieve` since the user is manually creating the string with the backslash. But any other method will be prone to fail.

For example, if we create a `registry.txt` file:
```
santisoler\/pooch-test-data-v1.zip md5:6cdda261f5646a4089966fd0bf505233
```
And we want to load it after creating a `Pooch` object:
```python
import pooch

pup = pooch.create(
    path=pooch.os_cache("testing-zenodo"),
    base_url="doi:10.5281/zenodo.7632643/",
    registry=None,
)
registry_file = "registry.txt"
pup.load_registry(registry_file)
```

Then the following `fetch` call will fail, since the file is loaded in a way that the backslash is interpreted as an escape character, complaining that the file doesn't exist in the registry:
```python
pup.fetch(r"santisoler\/pooch-test-data-v1.zip")
```

And this other call will also fail because the url will not be parse correctly:
```python
pup.fetch(r"santisoler/pooch-test-data-v1.zip")
```

Moreover, populating the registry through the repository API (see #319) will also fail at some point.

Is there any other way we can solve this considering that these filenames are only present in Zenodo?

santisoler:
I think this patch has a flawed design: having to introduce the backslash makes it more difficult to use by the end user, but also creates a lot of corner cases that are prone to future errors.

I came out with a better solution in #340, where the `parse_url` function has a special case for Zenodo dois. With this new design there's no need to introduce the backslash, making it to work flawlessly with both `retrieve` and a `Pooch` instance with a loaded registry file. Moreover, I suspect it will also work with the new feature to populate the registry in #319.

Therefore, I'm closing this PR.

PR #339, opened 2023-02-02
Port from deprecated appdirs to platformdirs

alexfikl:
This ports the uses (and mentions) of the deprecated `appdirs` to `platformdirs`. The APIs should be mostly equivalent (definitely for the ones used here) and `platformdirs` is already in use by most popular projects.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,7 +41,7 @@ packages = find:
 python_requires = >=3.7
 setup_requires =
 install_requires =
-    appdirs>=1.3.0
+    platformdirs>=2.5.0
     packaging>=20.0
     requests>=2.19.0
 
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -41,7 +41,7 @@ packages = find:
 python_requires = >=3.7
 setup_requires =
 install_requires =
-    appdirs>=1.3.0
+    platformdirs>=2.5.0
```
alexfikl, line 44: Is this min version? 2.5.0 was released a year ago, but older versions should work as well. This seems to be the bigger release:
https://github.com/platformdirs/platformdirs/releases/tag/2.1.0

santisoler:
Thanks @alexfikl for opening this PR! Very much appreciated! I think this is ready to be merged.

Regarding the minimum version of `platformdirs`, I think `v2.5.0` is a good floor.

PR #340, opened 2023-02-11
Support Zenodo urls with slashes in the filename

### LINKED ISSUE #336: error downloading file from sub-directory in zenodo repository

### LINKED ISSUE #341: Add test for `load_registry_from_doi` when the doi has a slash in the filename
**Description of the desired feature:**

We should add a test function that checks if the `load_registry_from_doi` method introduced in #325 will work with the bugfix introduced in #337. Probably would be better to create a utils function that returns the url to the special Zenodo repo that contains a slash in the filename and reuse it among the test functions.

This could be solve in #337 once #325 is merged. Otherwise, it's ok to leave it for its own PR.

**Are you willing to help implement and maintain this feature?**  For sure!

### LINKED ISSUE #366: Feature: allow null hash in registry files
**Description of the desired feature:**
[registry files](https://www.fatiando.org/pooch/latest/registry-files.html) are a nice way to load a large registry without a dict, and they also support custom urls for each file.  However, if one doesn't know the hash value up front and wants to [bypass the hash check](https://www.fatiando.org/pooch/latest/hashes.html#bypassing-the-hash-check), then you need to either use a dict instead of a registry file, manually or clear all the values after calling `load_registry`:

```python
for key, val in POOCH.registry.items():
    if val == sentinel:
        POOCH.registry[key] = None
```

**suggestion**: Since the registry file lines *must* have 3 items in order to use a custom url,  would it be possible to have a sentinel value like `None` or `_` or something to indicate that the hash should be ignored?

```txt
# registry.txt
myfile.png None https://www.example.com/m9zyiek2kr9pbed/myfile.png
```

santisoler:
Allow `DOIDownloader` to work with files uploaded to Zenodo that have a forward slash "/" in their name. These special cases are generated by the Zenodo-GitHub integration service, where the first portion of the filename matches the GitHub handle of the repository owned and the last portion matches the repository name. Now the `parse_url` function handles Zenodo dois under a special case: the netloc will be formed with the first two "parts" (`prefix/suffix`), while the path will be formed by everything that follows the second slash.

**Relevant issues/PRs:**

Fixes #336, fixes #341 and outperforms the idea drafted in #337

### DIFF
--- a/pooch/downloaders.py
+++ b/pooch/downloaders.py
@@ -622,7 +622,10 @@ class DOIDownloader:  # pylint: disable=too-few-public-methods
             )
 
         # Resolve the URL
-        file_name = parsed_url["path"].split("/")[-1]
+        file_name = parsed_url["path"]
+        # remove the leading slash in the path
+        if file_name[0] == "/":
+            file_name = file_name[1:]
         download_url = data_repository.download_url(file_name)
 
         # Instantiate the downloader object
--- a/pooch/tests/test_core.py
+++ b/pooch/tests/test_core.py
@@ -28,6 +28,7 @@ from .utils import (
     data_over_ftp,
     pooch_test_figshare_url,
     pooch_test_zenodo_url,
+    pooch_test_zenodo_with_slash_url,
     pooch_test_dataverse_url,
     pooch_test_registry,
     check_tiny_data,
@@ -41,6 +42,7 @@ REGISTRY = pooch_test_registry()
 BASEURL = pooch_test_url()
 FIGSHAREURL = pooch_test_figshare_url()
 ZENODOURL = pooch_test_zenodo_url()
+ZENODOURL_W_SLASH = pooch_test_zenodo_with_slash_url()
 DATAVERSEURL = pooch_test_dataverse_url()
 REGISTRY_CORRUPTED = {
     # The same data file but I changed the hash manually to a wrong one
@@ -632,6 +634,26 @@ def test_load_registry_from_doi(url):
         assert len(pup.registry) == 2
         assert "tiny-data.txt" in pup.registry
         assert "store.zip" in pup.registry
+
+        # Ensure that all files have correct checksums by fetching them
+        for filename in pup.registry:
+            pup.fetch(filename)
+
+
+def test_load_registry_from_doi_zenodo_with_slash():
+    """
+    Check that the registry is correctly populated from the Zenodo API when
+    the filename contains a slash
+    """
+    url = ZENODOURL_W_SLASH
+    with TemporaryDirectory() as local_store:
+        path = os.path.abspath(local_store)
+        pup = Pooch(path=path, base_url=url)
+        pup.load_registry_from_doi()
+
+        # Check the existence of all files in the registry
+        assert len(pup.registry) == 1
+        assert "santisoler/pooch-test-data-v1.zip" in pup.registry
 
         # Ensure that all files have correct checksums by fetching them
         for filename in pup.registry:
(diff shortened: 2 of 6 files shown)


santisoler:
@ djmannion, can you confirm that this other patch works locally for you? I think this a much better solution that the one before.

djmannion:
Yes, confirming that this version fixes my issue - and not having to specifically handle the extra slash is great from a user perspective, thanks!

santisoler:
Nice to hear that @djmannion. I'm merging this. I'm planning to make a Pooch release before the end of the month, so this bugfix will be available very soon.

PR #344, opened 2023-02-20
Automatically add trailing slash to base_url on pooch.create

### LINKED ISSUE #324: Automatically add a trailing "/" to base_url in pooch.create

In order to define a `Pooch` with a DOI, the DOI needs to end on a `/`. However, that character is not part of the DOI (e.g. https://dx.doi.org/ does not understand the DOI with `/`) and should not be needed here. The root of the problem is very likely a too similar handling of URLs and DOIs.

Using the example from [the official docs](https://www.fatiando.org/pooch/latest/protocols.html#digital-object-identifiers-dois) and stripping the trailing `/` from `base_url`:
```python
POOCH = pooch.create(
    path=pooch.os_cache("plumbus"),
    # Use the figshare DOI
    base_url="doi:10.6084/m9.figshare.14763051.v1",
    registry={
        "tiny-data.txt": "md5:70e2afd3fd7e336ae478b1e740a5f08e",
        "store.zip": "md5:7008231125631739b64720d1526619ae",
    },
)

def fetch_tiny_data():
    """
    Load the tiny data as a numpy array.
    """
    fname = POOCH.fetch("tiny-data.txt")
    data = numpy.loadtxt(fname)
    return data
```

**Full error message**

```
ValueError: Archive with doi:10.6084/m9.figshare.14763051.v1tiny-data.txt not found (see https://doi.org/10.6084/m9.figshare.14763051.v1tiny-data.txt). Is the DOI correct?
```

gabrielfu:
Automatically add a trailing "/" to base_url in `pooch.create()` if missing. Update docstring and add test function that checks if the trailing slash gets added.

**Relevant issues/PRs:**

Fixes #324

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -295,8 +295,8 @@ def create(
         Base URL for the remote data source. All requests will be made relative
         to this URL. The string should have a ``{version}`` formatting mark in
         it. We will call ``.format(version=version)`` on this string. If the
-        URL is a directory path, it must end in a ``'/'`` because we will not
-        include it.
+        URL does not end in a ``'/'``, a trailing ``'/'`` will be added
+        automatically.
     version : str or None
         The version string for your project. Should be PEP440 compatible. If
         None is given, will not attempt to format *base_url* and no subfolder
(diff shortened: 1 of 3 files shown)


santisoler:
Thanks for the contribution @gabrielfu! I think this is ready to be merged! I just added a small test function to prevent us breaking this feature in the future.

I'm merging it as soon as the Actions pass. 🚀

PR #348, opened 2023-02-27
Add support for Python 3.11

hugovk:
Add support for Python 3.11:

* Test it
* Add Trove classifer

Bump some CI checks from "3.10" to "3.x" (currently 3.11). Didn't do it for `.github/workflows/style.yml`, because that would require upgrading pylint to a later version to support 3.11 but also introduce failing lint rules.

Also bump GitHub Actions.

**Relevant issues/PRs:**

n/a


leouieda:
Thanks for this as always @hugovk. Merged in main here to see if the failing tests get fixed.

PR #355, opened 2023-03-13
Support Dataverse files without a persistentID

### LINKED ISSUE #354: Unable to download files in Dataverse repositories when files don't have PIDs

Hi! I was happy to hear that pooch has support for downloading files in repositories that use Dataverse. I played around with the commands a bit today.

The pooch.create and DOIDownloader functions work great when the files I want to download in Dataverse repositories have persistent identifiers. It looks like both functions assume that files in Dataverse repositories will have PIDs.

But many repositories using Dataverse don't register PIDs for their files (see [a conversation in the Dataverse Google Group](https://groups.google.com/u/1/g/dataverse-community/c/hFZRMXJ_-CE?hl=en) about which repositories do and don't). We can't assume that all files in Dataverse repositories will have PIDs, and it's likely that more repositories using Dataverse will "turn off" PID registration for their files.

So when I try to download a file that doesn't have a PID, I get an error.

Using pooch.create:
```python
data = pooch.create(
	base_url='doi:10.7910/DVN/DCDKZQ',
	path=pooch.os_cache('testing'))
data.load_registry_from_doi()
datafile = data.fetch(
	fname='Indata_2022.10.02_19.44.51.zip')
```

Using DOIDownloader:
```python
downloader = DOIDownloader()
url = 'doi:10.7910/DVN/DCDKZQ/Indata_2022.10.02_19.44.51.zip'
downloader(
	url=url, 
	output_file='/testing/Indata_2022.10.02_19.44.51.zip',
	pooch=None)
```

**Full error message**

```
  File "/testing_pooch.py", line 28, in <module>
    downloader(
  File "/usr/local/lib/python3.9/site-packages/pooch/downloaders.py", line 611, in __call__
    downloader(download_url, output_file, pooch)
  File "/usr/local/lib/python3.9/site-packages/pooch/downloaders.py", line 207, in __call__
    response.raise_for_status()
  File "/usr/local/lib/python3.9/site-packages/requests/models.py", line 943, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://dataverse.harvard.edu/api/access/datafile/:persistentId?persistentId=
[Finished in 2.9s with exit code 1]
```

**System information**

* Operating system: Mac OS 13.2.1
* Python installation (Anaconda, system, ETS): System (I think)
* Version of Python: 3.9.12
* Version of this package: v1.7.0

santisoler:
Add support for downloading Dataverse files that don't have a persistent ID. Use the file ID instead.

**Relevant issues/PRs:**

Fixes #354

**TODO**

- [ ] Add tests that download files that don't have a persistentID


santisoler:
We should add some tests for this bugfix that downloads some file from a Dataverse repository that doesn't provide persistent ID for its files. I would avoid using preexisting repositories, we want very small files since these tests will be run multiple times. Creating a version of the [test data for Pooch](https://doi.org/10.11588/data/TKCFEF) in another Dataverse repository would be nice.

pooch/downloaders.py
```diff
@@ -1034,20 +1034,31 @@ def download_url(self, file_name):
...
+            for file in response["data"]["latestVersion"]["files"]
+        }
+        if file_name not in files:
+            raise ValueError(
+                f"File '{file_name}' not found in data archive "
+                f"{self.archive_url} (doi:{self.doi})."
+            )
+        # Generate download_url using persistentId or file id
+        persistent_id = files[file_name]["persistentId"]
+        if persistent_id:
```
pdurbin, line 1050: The file ID will always be there. It's the primary key in the database.

The persistent ID (DOI or Handle) is optional so you can't rely on it being there. I would simply avoid even checking for it if all you want to do is download the file.

pdurbin (commented):
More details in the comment I left but I'd suggest always downloading files using the (database) ID.

Please feel free to hit me up on https://chat.dataverse.org if you have any questions. Thanks for teaching Pooch to download from Dataverse! 🐶 ❤️

pooch/downloaders.py
santisoler: I see. I thought the `persistentId` is always there, but it's empty if the files doesn't have one. Sorry for the misunderstanding.

If that's the case, you're right, we shouldn't assume that `persistentId` will always be there. I'll change the `if` statement then.

BTW, do you have any example where the `persistentId` is not even included in the response (I'm thinking for testing purposes)?

pdurbin: I would suggest changing your strategy to this:

- Only check for `id`

Simple! 😄

leouieda (commented):
Since the [API docs](https://guides.dataverse.org/en/5.13/api/dataaccess.html#basic-file-access) say:

> Basic access URI:
> /api/access/datafile/$id

and only after that they have a box saying you can also use the persistentID, I think we can go with @pdurbin's advise of only using the ID instead of PID. That would simplify the testing and code and it would only break if Dataverse were to break their API. That should be fine since if we assume they can break by removing the ID then they could break in so many other ways that we have no way to predict. If it does happen, we can always issue a patch. But I think it's unlikely that they will without bumping the API version. The case of Zenodo from #373 seems like a good example that things may break but unintentionally. In which case, we probably only have to report the issue.

leouieda:
I can merge in main and make the changes since @santisoler is on vacation.

leouieda:
Plus, finding a dataverse instance that doesn't have persistentIDs that would be willing to host the Pooch test data would be non-trivial. And they could always enable the PIDs and break our tests without any warning.

pdurbin:
15f7536 looks like a nice simplification.

By the way we have a new changelog for breaking changes to the Dataverse API, which we hope to keep short! Here's how it looks as of Dataverse 6.1: https://guides.dataverse.org/en/6.1/api/changelog.html

leouieda:
Thanks for sharing @pdurbin!

PR #359, opened 2023-03-25
Fix grammar for retrieve documentation

rob-luke:
Thank you for the lovely package. I noticed this small typo when reading the docs.


leouieda:
Thanks for that @rob-luke! Much appreciated ❤️

PR #361, opened 2023-05-22
CI: Only run cron for upstream

hugovk:
We only need to run the cron schedule on upstream, and can skip for forks.

More details: https://dev.to/hugovk/til-how-to-disable-cron-for-github-forks-2d0l


leouieda:
Good catch!

PR #365, opened 2023-06-17
Ensure processor has extracted all members when members=None

### LINKED ISSUE #364: If Unzip/Untar fails because of a non-existent member, calling without members won't work

Unfortunately #316 still doesn't quite fix the problem of dealing with members in archives. It creates a race condition where calling with members before calling without them will cause the `members=None` call to return an incomplete archive — and in the case of a failed earlier call, it will be empty.

The issue is that if members is None, we still only check for the existence of the unzip directory:

https://github.com/fatiando/pooch/blob/9c57dd38c8fa8d20e2c4f2cfa41ac3d33be78138/pooch/processors.py#L84-L86

But the directory does not necessarily contain every member we need.

(With thanks to @dokempf for the example 😃)

```python
import pooch

P = pooch.create(
    path=pooch.os_cache("test"),
    base_url="https://github.com/ssciwr/afwizard-test-data/releases/download/2022-06-09/",
    registry={
        "data.tar.gz": "sha256:fae90a3cf758e2346b81fa0e3b005f2914d059ca182202a1de8f627b1ec0c160"
    }
)
try:
    files1 = P.fetch("data.tar.gz", processor=pooch.Untar(members=["i don't exist"]))
except:
    pass
files2 = P.fetch("data.tar.gz", processor=pooch.Untar())

assert len(files2) > 0
```

The assertion fails, and will fail forever now, even if we remove the faulty fetch! (State is hard. 😅)

```
Downloading file 'data.tar.gz' from 'https://github.com/ssciwr/afwizard-test-data/releases/download/2022-06-09/data.tar.gz' to '/Users/jni/Library/Caches/test'.
Extracting 'i don't exist' from '/Users/jni/Library/Caches/test/data.tar.gz' to '/Users/jni/Library/Caches/test/data.tar.gz.untar'
  File "/Users/jni/projects/play/pooch-test.py", line 16, in <module>
    assert len(files2) > 0
AssertionError
```

**Suggested fix**

At the start of `__call__`, we should check if `members` is None, and if so, set `members` to the full list of files from the archive. Then I think everything else should work nicely. 😊 And it might simplify some of the later conditionals.

**System information**

* Operating system: macOS 13.4
* Python installation (Anaconda, system, ETS): conda
* Version of Python: 3.10
* Version of this package: 1.7.0
* If using conda, paste the output of `conda list` below: (nah you don't want this 😂)

jni:
Here's an early attempt at fixing #364. Currently, test_unpacking is failing
because of the log tests. I'm not 100% sure what to do about them because the
log becomes pretty complex, so I don't know how to generate the expected log.
Maybe this is the wrong approach — happy to revert the last commit and try
something else if you have suggestions! 🙏

## Commits

- Add failing test for #364
- Add _all_members method for when members=None

**Relevant issues/PRs:**

Fixes #364


jni:
👋

leouieda:
Hi @jni sorry for the very very long delay on this. We'll try to put out the fire on #373 and then get on this.

leouieda:
@jni I think the problem is that you're changing the value of `self.members`. There is a check in `_extract_file` for `self.members is None` to decide how to extract and that check is now always false. I think the fix would be to set a local `members` variable in `__call__` that either is `self._all_members(fname)` or `self.members`.

jni:
Thanks @leouieda! I'm sure I have older neglected PRs as a maintainer. 😂 And your suggestion seems to have done the trick! 🚀

pooch/tests/test_processors.py
```diff
@@ -255,3 +255,28 @@ def _unpacking_expected_paths_and_logs(archive, members, path, name):
...
+    https://github.com/fatiando/pooch/issues/364
+    """
+
+    with TemporaryDirectory() as local_store:
+        pup = Pooch(path=Path(local_store), base_url=BASEURL, registry=REGISTRY)
+
+        # Do a first fetch with incorrect member
+        processor1 = processor_class(members=["i don't exist"])
+        filenames1 = pup.fetch("store" + extension, processor=processor1)
+        assert len(filenames1) == 0
```
leouieda, line 277:
```suggestion
        processor1 = processor_class(members=["tiny-data.txt"])
        filenames1 = pup.fetch("store" + extension, processor=processor1)
        assert len(filenames1) == 1
```

@jni wouldn't it make more sense to test extracting a single file that does exist? This should still cause a problem with the old behavior right? Plus, using a bad file name should really cause an error (as you mentioned previously, probably best to open a separate issue 🙂).

leouieda (changes requested):
@jni small comment on the test and I think Black and pylint are complaining 😬 

Thanks for this!

leouieda (approved):

leouieda:
@jni sorry for the delay on this. I'll get these documentation issues fixed in main and then merge this asap. Thanks for your help!

jni:
@leouieda don't worry, I am very comfortable with long-lived PRs. 😂 I appreciate your efforts! 🙏 Please let me know if I can do anything else.

leouieda:
Alright! I made some slight modifications to make ExtratorProcessor an abstract base class. That got rid of the need to check if methods are implemented and made the tests and checks pass with our new pylint configuration.

Thanks for the contribution @jni! ❤️

PR #370, opened 2023-08-15
Only create local data directories if necessary

shoyer:
Creating a local data directory with Pooch checks that the local directory is writeable. This is important if downloading data, but if the files already exist is unnecessary.

This came up because we would like to store data for Pooch in readable but non-writable directories. In particular, the directories in which our build system installs Python are not writable.

### DIFF
--- a/pooch/core.py
+++ b/pooch/core.py
@@ -217,15 +217,17 @@ def retrieve(
         path = os_cache("pooch")
     if fname is None:
         fname = unique_file_name(url)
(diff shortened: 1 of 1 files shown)


santisoler:
Thanks @shoyer for opening this PR. It looks good to me! I would wait for @leouieda's review to merge it.

leouieda:
Thanks @shoyer and @santisoler!

Issue #330, opened 2022-12-03
Interest in adding type annotations?


rsokl:
Hello! I am a big fan of pooch! 😄 

Would you be interested in my adding type annotations to the library? I think that this can help improve the user experience -- via enhanced 'intellisense' in IDEs when working with pooch, and for instant feedback when a user makes a mistake. I have experience creating/maintaining [hydra-zen](https://github.com/mit-ll-responsible-ai/hydra-zen) which makes extensive use of advanced typing features, and I have helped [Hypothesis](https://github.com/HypothesisWorks/hypothesis) adopt type annotations.

Some things that this would entail:
- Adding `typing_extensions` as a dependency (this is already quite ubiquitous, and is practically part of the std-lib). This provides backports of nice typing features, such as `Literal` and `Protocol`, so that these features can be leveraged without breaking compatibility with older versions of Python.
- Adding both mypy and pyright CI jobs that scan of source and example code, in order to ensure that the annotations are consistent and accurate, and so that they remain this way.

Let me know what you think!

santisoler:
Hi @rsokl! Sorry for the delayed reply. I'm really glad you opened this issue. We have been talking about type annotations but we haven't decided yet if this is something we want to implement. I know that introducing them offers some benefits when working inside IDEs, is there any other benefit that you can tell us about?

On the other side I found a few drawbacks of type annotations:
- Lowers the readability of the code. When using type annotations the function definitions are a bit more complex. I think this impacts specially on new users and new contributors: understanding how to use or contribute to that code base could be harder if they aren't familiar with type annotations (and most Python novices aren't).
- It doesn't enforce type checking. We don't get any type checking on runtime by adding type annotations, only some IDEs and language servers use them to warn the user  if they are passing an "offending" value. Therefore the benefits of type annotations would be leveraged only by users of those specific IDEs and LS. Please correct me if I'm wrong.
- It won't replace the docstrings, so we are duplicating stuff. I don't think that type annotations should replace the type specifications we include on the docstrings of our functions, therefore adding type annotations would mean that we have duplicated specifications of the "accepted" types for each parameter. Is there a tool that can automatically check if type annotations and docstrings differ?
- It somewhat limits the duck-typing spirit of Python. Wether I like it or not, ducktyping is a thing in Python. Sometimes it bites us, but in a lot of situations we can leverage it. It seems that type annotations could impose a soft boundary on the type of the arguments we pass to our functions/methods. Am I right or do I have a wrong perception of this?
[...]

PR #404, opened 2024-03-11
Add mypy to CI job and type hints for one class.

eliotwrobson:
Changes to test type hints and integrating a typechecker into the CI job.
This was to test feasibility of adding type annotations to the package (per
the issue). This was surprisingly easy to do, I basically just copied the types
listed in the docstring.

**Relevant issues/PRs:**

https://github.com/fatiando/pooch/issues/330

**Squash and Merge summary**

Add type hints to `pooch/core.py` and create a new `typing` submodule for custom type classes, and add it to the API Reference. Run `mypy` on CI to perform type checks, and create new targets in the `Makefile`. Extend the list of dependencies required to run the type checks.


eliotwrobson:
Hi @santisoler, is there anything I could do to help move this along? This change would be helpful to me when working with this library, and it seems like there has been interest from others as well.

env/requirements-style.txt
```diff
@@ -4,3 +4,8 @@ flake8
 pylint>=2.4
 pathspec
 burocrata
+types-requests
+types-tqdm
+types-paramiko
+mypy
+pytest
```
santisoler, line 11: Is there a reason why we need `pytests` for checking styles?

pooch/core.py
```diff
@@ -27,6 +29,28 @@
...
+        self,
+        fname: str,
+        action: t.Optional[FilePath],
+        pooch: "Pooch",
+        *,
+        check_only: t.Optional[bool] = None,
+    ) -> t.Any: ...
+
+
+ProcessorT = t.Callable[[str, Actions, "Pooch"], t.Any]
```
santisoler, line 52: I think I'd like to separate the public classes from the type variables in different submodules. What about defining a `pooch.typing` submodule where these can live?

On a side note, would it be feasible to name them without the `T`? Or will it create more confusion?

santisoler (commented):
Hi @eliotwrobson. Yes, I totally agree. Type hints is something people would like to see in Pooch (me included!).

I left a few comments here so we can move forward with this PR.

## Importing from `typing`

I noticed that in this PR you `import typing as t`, and then use each one of the types in signatures with `t.<TYPE>`, like `t.Optional, `t.List`, etc.

Why not importing the required types directly? Like `from typing import ...`. Most of the codebases that use type hints use the modules like this:

```python
from typing import Optional

def f(x: Optional[float] = None):
    ...
```

I think having that extra `t.` creates some clutter in the signatures, specially when we need to use several types. What do you think?

## Do we need `typing_extensions`?

I noticed that you added `typing_extensions` as a dependency. I'm curious if this strictly necessary. I left a comment below where we can continue the conversation.

## Minor things to do

Before I forget, we need to add the new style requirements to `environment.yml` as well. Thanks for adding them to `requirements-style.txt` and to add a new target to the Makefile!

---

Thanks again for the effort you put in this! I really appreciate that we can move forward with type hints in Pooch!

pooch/core.py
eliotwrobson: I think that would work. The T is my convention but I can remove that as well.

env/requirements-style.txt
eliotwrobson: Mypy is run over the test case files as well, so this needs pytest installed to recognize the import.

eliotwrobson:
@santisoler I've responded to your comments. I've added mypy to the environment.yml, removed typing_extensions, and changed the typing import.

santisoler:
Hi @eliotwrobson! Thanks for addressing my comments, I appreciate it.

I've just pushed a few changes to your branch, hope you don't mind (feel free to revert changes if you disagree with them). I managed to move the new type classes to their own `pooch.typing` module. I solved the circular loop by conditionally importing `Pooch` only if `typing.TYPE_CHECKING` is `True`. I took the liberty to rename some of the names to reflect better their meaning. For example, `Actions` to `Action`, since the type represents a single `str` that matches a single action. I also renamed `FilePath` for `PathType`, since this type hint is usually used for directories, and not only files. I also added the new submodule to the API reference in the docs. Please, let me know what do you think about these changes.

I still see a few mypy errors. Would you like to work on them?

eliotwrobson:
> I've just pushed a few changes to your branch, hope you don't mind (feel free to revert changes if you disagree with them).

No worries, thank you for making these changes! Everything looks good to me. Would you like to add typing to everything in the library before merging this? Or merge as-is and incrementally add typing over a number of PRs? This currently doesn't have the `py.typed` file, so client code won't be able to see the annotations that have been added so far, which IMO is good before everything in the library has had types added.

santisoler:
I think it would be better to merge this PR and work on adding other type hints incrementally in other PRs. That would things easier for reviewing, for avoiding conflicts and to also see some progress.

I'm afraid I'm not familiar with the `py.typed` file, so if you think it's worth adding it while we work on it (instead of waiting for having everything type hinted), feel free to do so. I'll try to get more familiar with it in the meantime.

eliotwrobson:
@santisoler sounds good. I don't think it's worth adding `py.typed` until everything has been type annotated, since the presence of this file exposes the annotations to client code, and I don't think this is a good idea until annotations are on everything (which shouldn't take too long since this library isn't too huge).

Let me know if you'd like more changes here before merging. If not, I think everything looks good, so after this is merged I can get started on annotations in other files.

santisoler:
> I don't think it's worth adding py.typed until everything has been type annotated, since the presence of this file exposes the annotations to client code, and I don't think this is a good idea until annotations are on everything (which shouldn't take too long since this library isn't too huge).

I see. Sorry, I misunderstood your previous comment. Let's forget about the `py.typed` file for now then.

> Let me know if you'd like more changes here before merging. If not, I think everything looks good, so after this is merged I can get started on annotations in other files.

I think we are good to go. I just made a small change to `doc/conf.py`: the [`typing.List` is deprecated](https://docs.python.org/3/library/typing.html#typing.List), so I replaced it for the builtin `list` type.

I'm merging this! 🚀 Thanks again for pushing forward with this and for the hard work.

Feel free to open another PR for type hinting the rest of Pooch whenever you want, no rush!. And also, feel free to split the work in several PRs.

PR #408, opened 2024-03-18
Add projects using pooch

jlaehne:
Add HyperSpy, RosettaSciIO and eXSpy to the list of projects using pooch (for on demand loading of large test files)


leouieda:
Thanks, @jlaehne! I'm glad Pooch was useful for you.

PR #412, opened 2024-05-17
Mark additional tests requiring network

penguinpee:
Those tests also require network and should be excluded when running
`pytest -m 'not network'`.

**Summary for Squash and Merge**

Add `network` pytest mark to test functions that require network connection to hit third party servers. Improve `test_check_availability_invalid_downloader` to remove the need to use network connection just to test an error raised after an invalid argument to a custom downloader.

### DIFF
--- a/pooch/tests/test_core.py
+++ b/pooch/tests/test_core.py
@@ -525,12 +525,9 @@ def test_check_availability_invalid_downloader():
         return None
 
     pup = Pooch(path=DATA_DIR, base_url=BASEURL, registry=REGISTRY)
-    # First check that everything works without the custom downloader
-    assert pup.is_available("tiny-data.txt")
-    # Now use the bad one
-    with pytest.raises(NotImplementedError) as error:
+    msg = "does not support availability checks."
+    with pytest.raises(NotImplementedError, match=msg):
         pup.is_available("tiny-data.txt", downloader=downloader)
-        assert "does not support availability checks" in str(error)
 
 
 @pytest.mark.network
@@ -632,6 +629,7 @@ def test_stream_download(fname):
     [FIGSHAREURL, ZENODOURL, DATAVERSEURL],
     ids=["figshare", "zenodo", "dataverse"],
 )
+@pytest.mark.network
 def test_load_registry_from_doi(url):
     """Check that the registry is correctly populated from the API"""
     with TemporaryDirectory() as local_store:
@@ -649,6 +647,7 @@ def test_load_registry_from_doi(url):
             pup.fetch(filename)
 
 
+@pytest.mark.network
 def test_load_registry_from_doi_zenodo_with_slash():
     """
     Check that the registry is correctly populated from the Zenodo API when
--- a/pooch/tests/test_core.py
+++ b/pooch/tests/test_core.py
@@ -624,12 +624,12 @@ def test_stream_download(fname):
         check_tiny_data(str(destination))
 
 
+@pytest.mark.network
 @pytest.mark.parametrize(
     "url",
     [FIGSHAREURL, ZENODOURL, DATAVERSEURL],
     ids=["figshare", "zenodo", "dataverse"],
 )
-@pytest.mark.network
 def test_load_registry_from_doi(url):
     """Check that the registry is correctly populated from the API"""
     with TemporaryDirectory() as local_store:
--- a/pooch/tests/test_downloaders.py
+++ b/pooch/tests/test_downloaders.py
@@ -86,6 +86,7 @@ def test_unsupported_protocol():
         choose_downloader("doii:XXX/XXX/file")
 
 
+@pytest.mark.network
 def test_invalid_doi_repository():
     "Should fail if data repository is not supported"
     with pytest.raises(ValueError) as exc:
@@ -96,6 +97,7 @@ def test_invalid_doi_repository():
     assert "Invalid data repository 'joss.theoj.org'" in str(exc.value)
 
 
+@pytest.mark.network
 def test_doi_url_not_found():
     "Should fail if the DOI is not found"
     with pytest.raises(ValueError) as exc:
@@ -112,6 +114,7 @@ def test_doi_url_not_found():
     ],
     ids=["figshare", "zenodo", "dataverse"],
 )
+@pytest.mark.network
 def test_figshare_url_file_not_found(repository, doi):
     "Should fail if the file is not found in the archive"
     with pytest.raises(ValueError) as exc:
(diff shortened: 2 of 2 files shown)


penguinpee:
Background: In Fedora the build environment is not network enabled and downloads are prohibited during build. Yet, we like to run tests whenever possible. This allows us to simply pass `-m 'no network'` to `pytest` for running all tests that don't require network.

penguinpee:
I marked a few more tests, which I discovered being unmarked after updating to the latest release.

pooch/tests/test_core.py
```diff
@@ -499,6 +499,7 @@ def test_check_availability():
     assert not pup.is_available("not-a-real-data-file.txt")
 
 
+@pytest.mark.network
```
santisoler, line 502: I don't think we need to mark this test as `network`, since it's using the `ftpserver` fixture provided by [`pytest-localftpserver`](https://github.com/oz123/pytest-localftpserver), which runs a local ftp server for testing the `FTPDownloader`. Have you run in any problem trying to run this test without internet connection?
```suggestion
```

pooch/tests/test_core.py
```diff
@@ -517,6 +518,7 @@ def test_check_availability_on_ftp(ftpserver):
         assert not pup.is_available("doesnot_exist.zip", downloader=downloader)
 
 
+@pytest.mark.network
```
santisoler, line 521: This test should not require network connection. The only reason it does is because it tests if everything works without the custom downloader, but this is already covered by `test_check_availability`. So I think we should remove the `network` mark and update the test to something like:
```python
def test_check_availability_invalid_downloader():
    "Should raise an exception if the downloader doesn't support this"

    def downloader(url, output, pooch):  # pylint: disable=unused-argument
        "A downloader that doesn't support check_only"
        return None

    pup = Pooch(path=DATA_DIR, base_url=BASEURL, registry=REGISTRY)
    msg = "does not support availability checks."
    with pytest.raises(NotImplementedError, match=msg):
        pup.is_available("tiny-data.txt", downloader=downloader)
```

pooch/tests/test_downloaders.py
```diff
@@ -471,6 +475,7 @@ class TestZenodoAPISupport:
             ("invalid", invalid_api_response),
         ],
     )
+    @pytest.mark.network
```
santisoler, line 478: This test makes use of the `httpserver` pytest fixture provided by [`pytest-httpserver`](https://pypi.org/project/pytest_httpserver/), which runs a local http server for running this test. Therefore I don't think we need to add a `network` mark to it. Have you run in any problem trying to run this test without internet connection?
```suggestion
```

pooch/tests/test_downloaders.py
```diff
@@ -495,6 +500,7 @@ def test_api_version(self, httpserver, api_version, api_response):
         "api_version, api_response",
         [("legacy", legacy_api_response), ("new", new_api_response)],
     )
+    @pytest.mark.network
```
santisoler, line 503: Same case with this test, which also makes use of `httpserver`. Do you face any problem if you run this test without connection?
```suggestion
```

pooch/tests/test_downloaders.py
```diff
@@ -522,6 +528,7 @@ def test_download_url(self, httpserver, api_version, api_response):
         "api_response",
         [legacy_api_response, new_api_response],
     )
+    @pytest.mark.network
```
santisoler, line 531: Same case with this test, which also makes use of `httpserver`. Do you face any problem if you run this test without connection?
```suggestion
```

santisoler (changes requested):
Thanks @penguinpee for opening this PR! Thanks for catching those unmarked test functions.

I left a few comments below, nothing major. Let me know what do you think.

BTW, don't worry about the failing test in `macos`. Is not your fault. I'll fix that in a separate PR.

pooch/tests/test_core.py
penguinpee: We don't have `pytest-localftpserver` packaged for Fedora and it seems I included this solely going by the name of the test. It seems easy enough to package, though. I'll give that a shot and come back to you with a definitive answer.

pooch/tests/test_core.py
penguinpee: Currently the test fails with
```python
E       socket.gaierror: [Errno -3] Temporary failure in name resolution
```

I'm fine with changing the test in favor of marking it.

pooch/tests/test_downloaders.py
penguinpee: Similar answer regarding `pytest-localftpserver`. Small difference: I didn't include the required `pytest` plugin. I'll do that and test again together with FTP tests.

pooch/tests/test_downloaders.py
penguinpee: Same case, same answer. ;)

pooch/tests/test_downloaders.py
penguinpee: Dito.

penguinpee:
In a nutshell, I'll make sure all plugins are present and will test again. I suppose you are right regarding the tests relying on `localftpserver` and `httpserver`. I'll update this PR according to my findings. Apologies for being a bit hasty in my conclusions.

pooch/tests/test_core.py
santisoler: Good to know. Does my new proposal pass on your side?

pooch/tests/test_downloaders.py
santisoler: Cool. For reference, here's a list of all requirements for running the tests: https://github.com/fatiando/pooch/blob/main/env/requirements-test.txt

pooch/tests/test_core.py
penguinpee: Yes, it does. Thank you very much.

pooch/tests/test_downloaders.py
penguinpee: Yeah. I just discovered that myself. I've taken over maintenance of `pooch` in Fedora recently. The previous maintainer used the PyPI sdist, which lacks the `env/` dir. I've switched to GitHub tarballs now. Makes life a little easier (apart from having to fight `setuptools_scm` :wink:).

penguinpee:
This now works for me with `httpserver` and `localftpserver` using `-m 'not network'.

santisoler (approved):
Thanks for addressing my comments. And no need to apologize, the reason why these decorators shouldn't be added is subtle.

I just pushed two minor commits that move the decorators up in the list, just to improve style, nothing critical. I'm merging this.

PR #423, opened 2024-06-08
Exclude doc/ from wheel

penguinpee:
The automatic discovery adds doc/ to the wheel triggered by the
presence of `conf.py` in that directory.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,8 +45,9 @@ xxhash = ["xxhash>=1.4.3"]
 "Bug Tracker" = "https://github.com/fatiando/pooch/issues"
 "Source Code" = "https://github.com/fatiando/pooch"
 
-[tool.setuptools.packages]
-find = {}  # Scanning implicit namespaces is active by default
+[tool.setuptools.packages.find]
+# Scanning implicit namespaces is active by default
+exclude = ["doc*"]
 
 [tool.setuptools.package-data]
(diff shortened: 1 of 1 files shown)


penguinpee:
Alternatively, disabling namespaces also works: `find = {namespaces = false}` instead of `find {}`.

leouieda:
Thanks for that @penguinpee. We're discussing changing our packaging practices to remove all of these from our built packages here: https://github.com/fatiando/community/issues/154

carlosal1015:
Here https://archlinux.org/packages/extra/any/python-pooch/files same error
```
error: failed to commit transaction (conflicting files)
python-pooch: /usr/lib/python3.12/site-packages/doc/conf.py exists in filesystem (owned by python-codepy)
```

penguinpee:
> Here https://archlinux.org/packages/extra/any/python-pooch/files same error
> 
> ```
> error: failed to commit transaction (conflicting files)
> python-pooch: /usr/lib/python3.12/site-packages/doc/conf.py exists in filesystem (owned by python-codepy)
> ```

In your case the file is already claimed by `codepy`. I looked at the sdist from PyPI and `codepy` is making the same mistake by using `packages=find_namespace_packages()` in `setup.py`, which adds `doc` as a top level module with `conf.py` in it.

That should be reported to `codepy` upstream. For now you could apply this PR as a patch to `pooch` allowing you to get on with the installation.

carlosal1015:
I will do it, many thanks for the feedback.

PR #437, opened 2024-10-01
Check all available checksum algorithm in DataVerse registry population

### LINKED ISSUE #435: DOI registry assumes md5 hashing algorithm

```python
import pooch
example = pooch.create(
    path=pooch.os_cache("example"),
    base_url="doi:10.34894/5SOKTV",
)
example.load_registry_from_doi()```
**Full error message**

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[20], line 1
----> 1 example.load_registry_from_doi()

File /usr/local/lib/python3.11/site-packages/pooch/core.py:704, in Pooch.load_registry_from_doi(self)
    701 repository = doi_to_repository(doi)
    703 # Call registry population for this repository
--> 704 return repository.populate_registry(self)

File /usr/local/lib/python3.11/site-packages/pooch/downloaders.py:1162, in DataverseRepository.populate_registry(self, pooch)
   1151 """
   1152 Populate the registry using the data repository's API
   1153 
   (...)
   1157     The pooch instance that the registry will be added to.
   1158 """
   1160 for filedata in self.api_response.json()["data"]["latestVersion"]["files"]:
   1161     pooch.registry[filedata["dataFile"]["filename"]] = (
-> 1162         f"md5:{filedata['dataFile']['md5']}"
   1163     )

KeyError: 'md5'```

**System information**

* Operating system:
* Python installation (Anaconda, system, ETS):
* Version of Python:
* Version of this package:
* If using conda, paste the output of `conda list` below:

<details>
<summary>Output of conda list</summary>
<pre>

PASTE OUTPUT OF CONDA LIST HERE

</pre>
</details>

dokempf:
Fixes #435.


dokempf:
Failures are unrelated. Should I look into fixing them?

santisoler:
Hi @dokempf! Thanks for taking care of this.

I don't think this fix solves the issue: with it the `populate_registry` doesn't throws an error, but it doesn't build the registry either. For example, if you run this code you'll see that the registry is still empty:

```python
import pooch

example = pooch.create(
    path=pooch.os_cache("example"),
    base_url="doi:10.34894/5SOKTV",
)
example.load_registry_from_doi()
print(example.registry)
```
```
{}
```

After inspecting the content of `self.api_response.json()["data"]["latestVersion"]["files"][0]` (assigned to `filedata` in the `populate_registry` method), I see that the dictionary doesn't contain the `"md5"` key anymore. Instead the checksum data is stored inside another dictionary under the `"checksum"` key:
```
{
    "id": 425558,
    "persistentId": "",
    "filename": "README.md",
    "contentType": "text/markdown",
    "friendlyType": "Markdown Text",
    "filesize": 2324,
    "storageIdentifier": "file://190251cc731-ca82ca1d341b",
    "rootDataFileId": -1,
    "checksum": {"type": "SHA-1", "value": "0e4b27fd3d76c75c37303f47e25d74ef407d0752"},
    "tabularData": False,
    "creationDate": "2024-06-17",
    "publicationDate": "2024-06-28",
    "fileAccessRequest": True,
}
```

I suspect there might be a change in the DataVerse API and how they return information about the checksum.

Therefore I think the fix should check if the `filedata` has either a `"checksum"` key and get the info from there, or if it has keys like the one you included in the fix. Would be nice to find some docs on the DataVerse API to understand the changes they made (sorry, I didn't search for it yet).


On a side note, don't worry about the pylint complain, I'll fix it in another PR. We should replace it for some modern alternative (`ruff` or `flake8`) at some point.

santisoler:
My bad. I just saw that you added a link to the docs in the issue. Thanks for that!

In the sample response they added in https://guides.dataverse.org/en/latest/api/native-api.html#import-a-dataset-into-a-dataverse-collection there's the `"checksum"` key I mentioned.

So, my take would be:
* Maintain support for the "md5" key (in case DataVerse instances are using the old API).
* Add support for other algorithm keys: "sha256", "sha1", etc.
* Add support for the new API with the "checksum" key.
* Raise a verbose error in case we couldn't find a checksum in the API response. I think if we couldn't populate the registry we should raise an error instead of quietly exiting.

santisoler:
BTW, I merged #438. If you update this branch you won't find the pylint error.

santisoler:
Hi @dokempf. Sorry for the delay, these weeks were a little bit busy. 

I noticed that your implementation was raising an error when "md5" and "checksum" were not in the response. Even if "sha1" (for example) was there, the code would just raise an error because the first two if statements were false. I just pushed a change that would fix it.

I also pushed a commit that adds some tests. I think we should include tests for this feature because of two reasons:
1. To make sure that the implementation works as expected both for the old and new APIs.
2. To ensure that any future refactor of this piece of code supports the same behaviours.

I know it's tedious to write these tests, but sometimes they are they only way to ensure things work as expected.

I still would like to add some more things to the tests:
- [ ] In the fake response I should use "SHA-1" instead of "sha1" (the current behaviour).
- [ ] Add a test that checks if the error is being raised.

But at least for now, the following example works:

```python
import pooch

example = pooch.create(
    path=pooch.os_cache("example"),
    base_url="doi:10.34894/5SOKTV",
)
example.load_registry_from_doi()
print(example.registry)

example.fetch("README.md")
```

pooch/tests/test_downloaders.py
```diff
@@ -543,3 +543,150 @@ def test_populate_registry(self, httpserver, tmp_path, api_response):
...
         downloader.populate_registry(puppy)
         assert puppy.registry == {self.file_name: f"md5:{self.file_checksum}"}
+
+
+class MockOldResponse:
+    """
+    Mock request response to test checksum algorithm in DataverseRepository.
+
+    Use the old API where the checksum algorithm was listed as a key in the
+    file information.
```
santisoler, line 553:
```suggestion
```

pooch/tests/test_downloaders.py
```diff
@@ -543,3 +543,147 @@ def test_populate_registry(self, httpserver, tmp_path, api_response):
...
+        }
+        checksum = "0123456789abcdef"
+        if self._checksum_key:
+            file_dict = {
+                "label": "foobar.txt",
+                "dataFile": {
+                    "id": 12345,
+                    "filename": "foobar.txt",
+                    "checksum": {
+                        "type": self._algorithm.upper(),
```
santisoler, line 637: This should be "SHA-1", "SHA-256", etc. Not just "SHA1" and "SHA256".

PR #453, opened 2024-12-20
Replace `pylint`, `flake8` and `black` with `ruff`

santisoler:
Replace the current linters and auto-formatters we are using in Pooch for `ruff`.

> [!NOTE]
>
> This branch is intended to collect all changes to complete the replacement of the linters. We should not push to this branch, but open PRs against it instead.
>
> I'm planning to merge this branch as soon as we get ruff happy with Pooch. We can leave a few warnings hanging and ignore them temporarily, so we can tackle them later.
>
> When ready, this branch should be **MERGED** to `main` and not Squash and Merged, so we don't create a huge commit.
>
> **Update: ** if possible, rebase `main` in to `ruff-dev` instead of merging to update it, so we don't generate those `Merge ...` commits.

> [!IMPORTANT]
>
> Make sure that the format passes on CI on PRs targeted to this branch. It's expected that the style check will fail until we solve all warnings, but format should pass.


leouieda:
TODO: Remove the pylint comments to disable rules per file/line.

pooch/hashes.py
```diff
@@ -222,7 +220,7 @@ def make_registry(directory, output, recursive=True):
...
 
     with open(output, "w", encoding="utf-8") as outfile:
-        for fname, fhash in zip(files, hashes):
-            # Only use Unix separators for the registry so that we don't go
-            # insane dealing with file paths.
-            outfile.write("{} {}\n".format(fname.replace("\\", "/"), fhash))
+        outfile.writelines(
+            "{} {}\n".format(fname.replace("\\", "/"), fhash)
+            for fname, fhash in zip(files, hashes)
+        )
```
santisoler, line 228:
```suggestion
        # Only use Unix separators for the registry so that we don't go
        # insane dealing with file paths.
        outfile.writelines(
            "{} {}\n".format(fname.replace("\\", "/"), fhash)
            for fname, fhash in zip(files, hashes)
        )
```

santisoler (commented):
Went through all files changed and all seems good to go.

santisoler:
@leouieda, if you have some time to check this out, I'd appreciate!

I'm considering squashing and merging this PR, against my initial idea described in the PR description. It ended up having quite a bunch commits, a significant number of them being just merge commits from `main`.

What if we just squash and merge this? We would still have the "history" through the PRs. And honestly, there are only a few changes that are not self-explanatory. Let me know what do you think!

leouieda:
@santisoler this looks good to me. I'm fine with squash merging this since it's all one big change anyway.

PR #462, opened 2025-02-25
Add remaining type hints to library

### LINKED ISSUE #330: Interest in adding type annotations?

eliotwrobson:
Add more type annotations and `py.typed` file.

**Relevant issues/PRs:**
Resolves #330, continuation of #404

### DIFF
--- a/pooch/__init__.py
+++ b/pooch/__init__.py
@@ -19,7 +19,7 @@ from .downloaders import (
 from .processors import Unzip, Untar, Decompress
 
 # This file is generated automatically by setuptools_scm
-from . import _version  # type: ignore
+from . import _version  # type: ignore[attr-defined]
 
 
 # Add a "v" to the version number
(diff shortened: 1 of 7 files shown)


flying-sheep:
can this be picked up?

Issue #456, opened 2024-12-29
More informative error message when resolving DOI URLs


remrama:
Hi, we have an open PR for the [YASA](https://raphaelvallat.com/yasa) library at raphaelvallat/yasa#192 that transitions from test files being shipped with our repository to using Pooch to download them. Everything is working well and all tests are now passing (Python versions 3.8-3.12 on ubuntu, macos, and windows). However, the initial tests failed on 2 of the 12 systems (3.11 and 3.12 macos). The failing tests passed when we reran them after 24 hours _without changing any code_, so I presume it was a temporary connection- or server-related issue.

The error raised during tests was from [L648-655](https://github.com/fatiando/pooch/blob/1bc1c51d6f6c70cf4b60fa35583b6429a7f5384b/pooch/downloaders.py#L648-L654) in `downloaders.py`, where Pooch tries to resolve the DOI, despite the DOI existing and passing in other tests.

```
ValueError: 'Archive with doi:10.5281/zenodo.14564284 not found (see https://zenodo.org/doi/10.5281/zenodo.14564284). Is the DOI correct?'
```

I have 2 questions:

1. Have you had similar experiences when running CI tests with Pooch downloads? If so, do you have any general advice on how best to handle potential connection issues when downloading files during CI tests (e.g., [caching](https://github.com/actions/cache))?
2. Would it be possible to make that `DOIDownloader` error message more informative? Currently it clusters all status codes 400-600 into the assumption that the DOI was wrong. I think it would be helpful if the response code itself was provided in the error message to help the user determine why the DOI was not resolved (e.g., client vs server issues). I would be willing to submit a small PR if this makes sense to add.

ophrys97:
I encountered the same issue. I had to bypass the DOI and download the files manually. However, I'm not sure if this is a temporary issue related to the server or connection. I will try again tomorrow to confirm. I hope the developers will address and resolve this issue as soon as possible.

santisoler:
Hi @remrama. Thanks for opening this issue.

It's not rare to encounter failed downloads when running tests on CI. Most of the times this is related to server-side issues, usually because we trigger multiple downloads in parallel. My pragmatic approach is to split jobs in Actions so I can rerun the failed ones without having to rerun all the tests. Another approach would be to include some sleeps in between jobs, so they are not triggered at the same time, although tests will take a little bit longer to run (I haven't experimented with that).

[...]

PR #477, opened 2025-08-05
Provide more descriptive errors when DOI request fails

santisoler:
Raise the `requests` response to provide more informative errors when the status code is between 400 and 600.

**Relevant issues/PRs:**

Part of #456

### DIFF
--- a/pooch/downloaders.py
+++ b/pooch/downloaders.py
@@ -651,9 +651,7 @@ def doi_to_url(doi):
     response = requests.get(f"https://doi.org/{doi}", timeout=DEFAULT_TIMEOUT)
     url = response.url
     if 400 <= response.status_code < 600:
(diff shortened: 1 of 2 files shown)


santisoler:
@leouieda, apparently the tests failures are due to figshare blocking some requests, particularly the ones ran from the macos hosts.

I'm ok if you want to skip these tests in order to make the release, or to just run them in a few hosts only. We can work this out later.

ericpre:
This would be useful to get this PR merged to diagnose issues caused by recently introduced rate limits in zenodo.

santisoler:
Hi @atravitz. Please, check #341 for more details on the limiting rates that Zenodo imposes and how the patch I added today in #507 might help to fix that.

PR #505, opened 2026-01-22
Extend support for Python 3.14

MridulS:
This adds support for 3.14 in the Trove classifiers and runs CI on 3.14  

**Relevant issues/PRs:**

Used this one as an example https://github.com/fatiando/pooch/pull/451

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ jobs:
   #############################################################################
   # Run tests
   test:
-    name: ${{ matrix.os }} python=${{ matrix.python }} dependencies=${{ matrix.dependencies }}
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -47,11 +47,11 @@ jobs:
...
           - dependencies: latest
-            python: "3.13"
+            python: "3.14"
           - dependencies: optional
-            python: "3.13"
-          # test on macos-13 (x86) using oldest dependencies and python 3.7
-          - os: macos-13
+            python: "3.14"
+          # test on macos-15 (x86) using oldest dependencies and python 3.9
+          - os: macos-15-intel
```
MridulS: macos-13 has been deprecated from github runner images.

MridulS:
I believe there are other open PRs that should fix the CI failures like https://github.com/fatiando/pooch/pull/484 so I guess I'll just let the CI fail here.

PR #540, opened 2026-06-02
Guard optional bz2/gzip/lzma imports so pooch imports without them (fixes #468)

### LINKED ISSUE #468: `lzma` import failed in Python 3.13

Issue is discussed here: https://github.com/pyspextool/pyspextool/issues/264

We're having a trouble with `lzma` import but only in Python 3.13.

```
File "/home/aburgasser/.local/lib/python3.13/site-packages/pooch/processors.py", line 15, in
import lzma
File "/usr/local/lib/python3.13/lzma.py", line 27, in
from _lzma import *
ModuleNotFoundError: No module named '_lzma'
```

**System information**

Python3.13 on linux, no conda environment

pooch is up to date (v2.3)

gaoflow:
Fixes #468

## Problem

`bz2`, `gzip` and `lzma` are *optional* features of the Python standard
library — an interpreter can be built without them when the corresponding
`bzip2`/`lzma`/`zlib` development headers are missing (reported on a
self-compiled Python 3.13 and on a custom Python 3.9 on RHEL in #468).

`pooch/processors.py` imported all three at module level:

```python
import bz2
import gzip
import lzma
```

so importing **any** part of pooch failed entirely on such interpreters:

```
File ".../pooch/processors.py", line 15, in <module>
    import lzma
ModuleNotFoundError: No module named '_lzma'
```

even for users who never touch `Decompress`.

## Change

As suggested in the issue, guard the three imports so a missing module
becomes `None` instead of crashing at import time:

```python
try:
    import lzma
except ImportError:
    lzma = None
```

The module is then only required when the matching `Decompress` method is
actually used. `Decompress._compression_module` now raises a clear
`ValueError` naming the missing module instead of failing later with an
`AttributeError` on `None`:

> Could not decompress 'data.xz' because the 'lzma' module is not available
> in this Python installation. This usually means Python was built without
> 'lzma' support. ...

When the modules are present (the normal case) behaviour is unchanged.

## Tests

* `test_decompress_unavailable_module` — monkeypatches the module table to
  `None` (simulating a Python built without the module) and checks the clear
  error for the `lzma`/`xz`/`bzip2`/`auto` paths.
* `test_processors_import_without_optional_modules` — spawns a subprocess
  that blocks importing `lzma`/`bz2` and asserts `import pooch.processors`
  still succeeds (faithful reproduction of #468).

Both fail on `main` and pass with this change; the full
`test_processors.py` suite (44 tests) passes and `ruff` is clean.

### DIFF
--- a/pooch/processors.py
+++ b/pooch/processors.py
@@ -10,15 +10,31 @@ Post-processing hooks
 """
 
 import abc
-import bz2
-import gzip
-import lzma
 import os
 import shutil
 import sys
 import typing
 from tarfile import TarFile
 from zipfile import ZipFile
+
+# bz2, gzip and lzma are optional features of the Python standard library: a
+# Python interpreter can be built without them (e.g. when the bzip2/lzma/zlib
+# development headers are missing). Importing them at module level would make
+# *any* use of pooch fail on such interpreters, even when no decompression is
+# needed. Guard the imports so the modules are only required when the matching
+# Decompress method is actually used (see GH #468).
+try:
+    import bz2
+except ImportError:
+    bz2 = None
+try:
+    import gzip
+except ImportError:
+    gzip = None
+try:
+    import lzma
+except ImportError:
+    lzma = None
 
 from .utils import get_logger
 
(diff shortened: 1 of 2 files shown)


binggao1230:
Pushed coverage follow-ups in `dc94dde` and `7ab6554`:

- Added the missing `gzip` unavailable-module cases for `Decompress._compression_module`.
- Changed the optional-module import test from a subprocess to same-process `importlib.reload`, so coverage records the `try` / `except ImportError` fallback lines in `pooch/processors.py`.

Local verification:

- `python -m pytest pooch/tests/test_processors.py -q` -> 46 passed
- `python -m pytest --cov=pooch --cov-config=.coveragerc --cov-report=term-missing pooch/tests/test_processors.py -q` -> 46 passed; `pooch/processors.py` at 100%
- `python -m ruff check pooch/tests/test_processors.py` -> all checks passed

Remote CI is now green, including `codecov/patch` and `codecov/project`.

PR #544, opened 2026-06-11
fix: apply tar "data" extraction filter whenever available (path traversal in Untar)

### LINKED ISSUE #543: Pooch Untar processor skips tarfile data filter on Python 3.9–3.11, allowing symlink path traversal outside the extraction directory
### Summary

`pooch.processors.Untar` conditionally passes `filter="data"` to `tarfile.extractall()` only when `sys.version_info >= (3, 12)`. On Python 3.9, 3.10, and 3.11 it calls `extractall()` with no tar extraction filter and no per-member path validation.

A malicious tar served from a caller-influenceable URL can contain a symlink entry pointing outside the extraction directory followed by a regular file entry written through that symlink. When a victim application calls `pooch.retrieve(url, known_hash=None, processor=Untar())` on an affected Python version, the downloaded tar is automatically extracted and the regular file is written outside the `.untar` destination.

This is an archive-extraction symlink path traversal / arbitrary file write issue in the `Untar` processor. It is bounded to Python versions where Pooch does not pass `filter="data"` to `tarfile.extractall()`. The attached PoC demonstrates the issue on Python 3.11.15 with Pooch 1.9.0 and shows that the same input is blocked on Python 3.12.3.

### Details

Affected component:

- repository: `fatiando/pooch`
- package: `pooch`
- file: `pooch/processors.py`
- affected processor: `Untar`
- affected function: `Untar._extract_file`

The vulnerable logic is the version-gated tar extraction filter:

```python
filter_kwarg = {} if sys.version_info < (3, 12) else {"filter": "data"}
...
tar_file.extractall(path=extract_dir, **filter_kwarg)
```

On Python versions below 3.12, `filter_kwarg` is `{}` and `tar_file.extractall()` receives no filter and no equivalent member validation. Therefore, tar entries with unsafe symlink targets are processed by the default unsafe tar extraction behavior on those Python versions.

The PoC uses two tar entries:

```text
1. symlink: sl -> <absolute path outside the .untar destination>
2. regular file: sl/owned.txt
```

During extraction, the first entry creates the symlink. The second entry is then written through the symlink, causing `owned.txt` to appear outside the extraction destination.

The bug is not simply that Python 3.11 is old. The attached live evidence records that Python 3.11.15 has `tarfile.data_filter` available, but Pooch's current `< (3, 12)` gate still skips passing `filter="data"` on that interpreter. The recommended fix is feature detection, not minor-version gating: pass `filter="data"` whenever the running `tarfile` supports it, or implement explicit per-member validation that rejects absolute paths, `..` traversal, and unsafe symlink/hardlink targets before extraction.

gaoflow:
## Problem

`Untar` enables the tarfile `data` extraction filter only on Python ≥ 3.12:

```python
filter_kwarg = {} if sys.version_info < (3, 12) else {"filter": "data"}
```

But the `data` filter (PEP 706) was backported to **3.9.17, 3.10.12 and 3.11.4**. On those interpreters Pooch skips the filter even though it is available, so `tarfile.extractall()` runs with no filter and no per-member validation.

A malicious archive served from a caller-influenceable URL can then escape the destination: a symlink member pointing outside `.untar`, followed by a regular file written *through* that symlink, writes the file outside the extraction directory when an application calls `pooch.retrieve(url, known_hash=None, processor=Untar())`. Reported in #543.

I verified this on Python 3.11.11 (which does have `tarfile.data_filter`): the two-entry PoC writes outside the destination under the current `< (3, 12)` gate, and is blocked (`AbsoluteLinkError`) once the filter is applied.

## Fix

Detect the filter by feature instead of minor version:

```python
filter_kwarg = {"filter": "data"} if hasattr(tarfile, "data_filter") else {}
```

The protection now applies on every interpreter that ships the filter, while interpreters without it keep the previous behavior (no regression). Scope is limited to `Untar`; `Unzip` relies on `ZipFile.extractall`, which already sanitizes `..` / absolute member names.

## Test

Adds `test_untar_blocks_symlink_path_traversal`: it builds the symlink-escape tar in memory, then asserts extraction raises `tarfile.FilterError` and that no file escapes the destination. The test is `skipif`-guarded on the rare interpreter without `data_filter`. It fails on the old `< (3, 12)` gate (the escape succeeds) and passes with the fix.

Closes #543

### DIFF
--- a/pooch/processors.py
+++ b/pooch/processors.py
@@ -15,7 +15,7 @@ import gzip
 import lzma
 import os
 import shutil
-import sys
+import tarfile
 import typing
 from tarfile import TarFile
 from zipfile import ZipFile
@@ -256,7 +256,13 @@ class Untar(ExtractorProcessor):
         This method receives an argument for the archive to extract and the
         destination path.
         """
-        filter_kwarg = {} if sys.version_info < (3, 12) else {"filter": "data"}
+        # Extract with the "data" filter to reject unsafe members (symlinks,
+        # hardlinks, or absolute/``..`` paths pointing outside the destination)
+        # that would otherwise let a malicious archive write files anywhere on
+        # disk. The filter was added in Python 3.12 and backported to 3.9.17,
+        # 3.10.12, and 3.11.4, so detect it by feature instead of minor version
+        # to protect every interpreter that supports it (see GH #543).
+        filter_kwarg = {"filter": "data"} if hasattr(tarfile, "data_filter") else {}
         with TarFile.open(fname, "r") as tar_file:
             if self.members is None:
                 get_logger().info(
--- a/pooch/tests/test_processors.py
+++ b/pooch/tests/test_processors.py
@@ -8,7 +8,10 @@
 Test the processor hooks
 """
 
+import io
+import os
 import re
+import tarfile
 from pathlib import Path
 from tempfile import TemporaryDirectory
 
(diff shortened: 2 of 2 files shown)


binggao1230:
Heads-up: the red CI here isn't from this change — every job fails in the "Collect requirements" step with `Could not find a version that satisfies the requirement dependente==0.3.0 (from versions: none)`, before any of the package or tests run. `dependente 0.3.0` is in fact the current release on PyPI, so this looks like a transient index hiccup during that run; a re-run should clear it (I don't have rerun rights on the repo).

PR #553, opened 2026-09-09
Avoid resource warnings

larsoner:
I was getting a bunch of `ResourceWarning`s in [MNE-Python CIs](https://app.circleci.com/pipelines/gh/mne-tools/mne-python/34287/details?useNewPipelines=true&workflowId=3d27f66d-a9e7-487c-95d3-ab06eaa18749&job=0d061cfa-71cf-41bb-8687-bfe09a440805&buildNumber=87804&jobType=build#step-172-):
```
ResourceWarning: unclosed <ssl.SSLSocket fd=14, family=2, type=1, proto=6, laddr=('30.0.158.241', 47702), raddr=('18.25.8.254', 443)>
```
Fix it by wrapping in context managers that close the resource.

The diff is really just a couple of lines plus an indentation change, so viewing with whitespace off might aid review.

This is really arguably an upstream bug with `urllib3-future` where a session can be closed too early, but by creating and using the `session`, `pooch` can avoid it.

**Relevant issues/PRs:**

None that I saw with a quick search


leouieda:
Thanks for reporting and for the fix! The test failure seems to be non-related (I'm rerunning to check) and the style failure as well. I'll take a closer look later and merge if nothing comes up.

PR #554, opened 2026-09-19
Use exist_ok when creating directory in stream_download

wpbonelli:
Avoid `FileExistsError` when `stream_download` is called by multiple processes

**Relevant issues/PRs:**
Same idea as #171 just a different code path


wpbonelli:
Looks like style checks are failing on main too, unrelated to these changes.

